PR #40, opened 2015-09-28
Fix missing spaces between key and value

matteblair:
The YAML spec requires key-value pairs to be separated with a colon and a space (except for JSON-like mappings with quotes). Some parsers and validators seem to tolerate a missing space, but it is invalid to the parser in tangram-es.


nvkelso:
Good catch! I've noticed funk around these types of properties before but never narrowed it down.

PR #44, opened 2015-09-29
Small fix

tallytalwar:
- visible is either a style property or a layer property


nvkelso:
@tallytalwar In this case, the line should be removed completely since the text: { visible: false } already covers this?

tallytalwar:
I did not know the intent, and the original placement (wrong) of `visible: false` was breaking yaml parsing on es (as its not a documented behavior)? 

All of pesky layer including sublayers are by default not visible. But that being said, since there are no sublayers, yes there is no need for this line. Will remove it and update the PR.

Edit: Updated and Rebased.

nvkelso:
LGTM, thanks for finding this!

Issue #26, opened 2015-09-23
Style buildings within landuse university, park and other green areas


sensescape:
- [x] Tiles has the data
- [x] Functional styling in scene file
- [x] Polish styling

<img width="367" alt="screen shot 2015-09-23 at 10 47 35 am" src="https://cloud.githubusercontent.com/assets/466585/10053421/c9f46832-61e0-11e5-9604-d64e3547953b.png">

sensescape:
Also within the shader

sensescape:
Need help with building shader to make it easier to set colors for building sides/facades
https://github.com/tangrams/eraser-map/blob/gh-pages/eraser-map.yaml#L775-L804

sensescape:
<img width="404" alt="screen shot 2015-09-29 at 11 44 16 am" src="https://cloud.githubusercontent.com/assets/466585/10174592/72024996-669f-11e5-9257-57feb8c302dd.png">

patriciogonzalezvivo:
a more comprehensive way of writing that is:

``` glsl
                    if ( dot(vec3(0.,0.,1.),normal) == 1.0 ) {
                        // is a roof
                        color.rgb = vec3(0.,0.,1.);
                    } else {
                        // is a wall
                        color.rgb = vec3(1.0,0.,0.);
                    }
```

At http://tangrams.github.io/tangram-play/?style=https://raw.githubusercontent.com/tangrams/eraser-map/gh-pages/eraser-map.yaml&lines=783-786#16.42333333333333/40.76963/-73.98839

![screen shot 2015-09-29 at 2 53 27 pm](https://cloud.githubusercontent.com/assets/346914/10174878/facf5a88-66b9-11e5-8860-029933a599ff.png)

sensescape:
fixed!

<img width="393" alt="screen shot 2015-09-29 at 3 58 16 pm" src="https://cloud.githubusercontent.com/assets/466585/10180565/149991be-66c3-11e5-8def-a18f2d188a0b.png">

sensescape:
with new roof colors:
<img width="258" alt="screen shot 2015-09-30 at 3 57 34 pm" src="https://cloud.githubusercontent.com/assets/466585/10208973/17009794-678c-11e5-9d45-c340b6083f96.png">
[...]

Issue #16, opened 2015-09-17
Buildings draw twice at some zooms


sensescape:
- [x] Just do it

https://github.com/tangrams/eraser-map/blob/gh-pages/eraser-map.yaml#L1380-L1440

-I'd like buildings footprints to appear at z14 for those buildings that extrude at z15 (can't get the filters at building-early to do this)

nvkelso:
@bcamper and @patriciogonzalezvivo for help optimizing the YAML filters.

bcamper:
No problem. To clarify, is this the desired logic:
- at z14, a subset of buildings render as footprint
- at z15, that same subset from z14 is now extruded, and all remaining buildings are drawn as footprint
- at z16, all buildings are extruded

Is that right?

bcamper:
Hm, looks like not all buildings are extruded until z17, but I still can't tell entirely what is supposed to happen, so please clarify :)

sensescape:
z14 - Yes, at z14 as a transition to the buildings that will get extruded at z15, their footprints show.
z15 - bldg footprints with height: { min: 190 } get extruded, while all buildings footprints with  area: { min: 500 } show
z16 - footprints with area: { min: 100 } show and buildings with height: { min: 100 } get extruded

These filters are being used for extrusion:
https://github.com/tangrams/eraser-map/blob/gh-pages/eraser-map.yaml#L1973-L1978

Peter's building extrusion transition with the tilting of the z-axis is also working here, so that short buildings extrude but don't look too short. And if they're very short, they stay as building footprints. I think this transition looks good : )

I think the main thing here is that the buildings footprints may be drawing twice since there are 2 building layers: buildings_early and buildings
And Nathaniel said the footprints are drawing twice. I didn't know how to clean it up and have it only draw once without breaking the current look now, which is good.

sensescape:
This is the building extrusion code from Peter:
https://github.com/tangrams/eraser-map/blob/gh-pages/eraser-map.yaml#L757-L767

nvkelso:
There’s duplication: the buildings that are extruded should be extruded,
but right now they also draw as **not** extruded first, then another pass
for extruded over them.

bcamper:
Working on this here: https://github.com/tangrams/eraser-map/compare/simplify-buildings

bcamper:
Addressed by PR #50.

bcamper:
@nvkelso @sensescape moved this up to priority 2 because of performance concerns.

[...]

PR #50, opened 2015-10-02
Simplify buildings

bcamper:
Remove building duplicate drawing, and reduce number of unique styles by consolidating color calculation logic. To address #16.

Because the diff is difficult to review, here's a walkthrough of the changes:
- Re-structured the draw rules for buildings.
  - Previously there were two layers, `buildings-early` (for footprints) and `buildings` (for extrusions), which caused substantial duplication in drawing of buildings, because it would have been difficult (or at least verbose) to write the rules in such a way as to avoid any overlapping between the two. 
  - Instead, all buildings are now drawn from a single `buildings` layer.
    - This layer first sets up the [default draw properties](https://github.com/tangrams/eraser-map/blob/simplify-buildings/eraser-map.yaml#L1880) for building fills and outlines, but sets `visible: false` so that no buildings are actually drawn until activated by later rules.
    - The [`footprints`](https://github.com/tangrams/eraser-map/blob/simplify-buildings/eraser-map.yaml#L1898) sub-layer turns on building visibility at different zoom, area, and height thresholds.
    - The [`extrude`](https://github.com/tangrams/eraser-map/blob/simplify-buildings/eraser-map.yaml#L1915) sub-layer then further turns on extrusion for a second, more restrictive set of criteria. Because `footprints` and `extrude` are sub-layers of the same ancestor layer, and both set properties on the same `polygons` and `lines` draw groups, they do not cause any duplication or conflict. When only `footprints` rules are met, a regular flat fill is used. When `extrude` criteria is also met, `extrude` properties are turned on.
  - The `extrude` sub-layer also sets the draw groups to draw with different styles: [`building-grid`](https://github.com/tangrams/eraser-map/blob/simplify-buildings/eraser-map.yaml#L1927) for the extruded volume, and [`building-line`](https://github.com/tangrams/eraser-map/blob/simplify-buildings/eraser-map.yaml#L1932) for the outline. These styles are described below.
  - Landuse-specific properties are then added with [`in_park`](https://github.com/tangrams/eraser-map/blob/simplify-buildings/eraser-map.yaml#L1936), `in_university`, etc. A consequence of these rules being consolidated in one layer is that landuse-specific coloring is now used for footprints as well as extrusions, as discussed in #26.
  - No changes to the actual filter criteria for footprints, extrusion, or landuse coloring.
- Building draw styles have been consolidated into a single `building-grid` style.
  - Previously, we had several styles with very slight variations (only for fixed colors), with `buildings_grid` as the base, then customized with styles such as `buildings_grid-parks`, `buildings_grid-hospital`, etc. It's fairly wasteful to compile and draw with many styles that have such slight variation.
  - To consolidate, I first moved the logic for scaling building height into its own style, [`scale-buildings`](https://github.com/tangrams/eraser-map/blob/simplify-buildings/eraser-map.yaml#L761). This removed copy-paste redundancy between the old `buildings_grid` and `buildings_lines` styles; the new versions of those styles now each [mix in the `scale-buildings` style](https://github.com/tangrams/eraser-map/blob/simplify-buildings/eraser-map.yaml#L776) so the code is shared.
  - The old `buildings_grid` and its derivatives had fixed color values for the roof and walls of each landuse type. To remove the extra derivative styles, I changed the style to use the built-in feature color (set in `layers` draw rules) as the roof color, and used an HSV calculation (modifying the saturation and value of the feature color) to derive the wall color (the [RGB is converted to HSV](https://github.com/tangrams/eraser-map/blob/simplify-buildings/eraser-map.yaml#L794), modified, then converted back to RGB for output). By tuning the [HSV params](https://github.com/tangrams/eraser-map/blob/simplify-buildings/eraser-map.yaml#L782), I was able to keep the visual very close. We can continue to tweak the color params as well, and if we find the approach too limiting, I have kept the older style derivatives [commented out](https://github.com/tangrams/eraser-map/blob/simplify-buildings/eraser-map.yaml#L803). Here's a comparison.
    - **Old (ignore labels):** screen shot 2015-10-03 at 10 28 06 pm **New:** screen shot 2015-10-03 at 10 27 37 pm
    - As with `scale-buildings`, a similar "helper" style was also added for HSV calculations, [`hsv`](https://github.com/tangrams/eraser-map/blob/simplify-buildings/eraser-map.yaml#L739). It is mixed into the [`building-grid`](https://github.com/tangrams/eraser-map/blob/simplify-buildings/eraser-map.yaml#L776) style.
  - `building-lines`, which is used to draw the outline around the footprint or extruded buildings, now becomes [very concise](https://github.com/tangrams/eraser-map/blob/simplify-buildings/eraser-map.yaml#L831) by reusing styling code from above.


nvkelso:
I'll have a look at this next week, too much code churn – needs a visual before and after session (or screenshots attached).

bcamper:
It should be almost exactly the same visual, except it adds the building
landuse colors from the extrusions to the footprints as well (because it
was easier to organize that way and @senescape said she was planning to add
them anyway). Should be way more efficient.

Don't let it get too far from master or it will be hard to merge...

bcamper:
Updated the PR description with walkthrough and screenshot side by side.

bcamper:
Updated against the recently rebased style. Visually re-verified changes. Merging to avoid further divergence. Building filter behavior is as described in #16, so should review be reviewed in case we want further tweaks.

nvkelso:
Thank you, looks good :)

PR #52, opened 2015-10-19
use gif for building-grid texture

meetar:


eraser-map.yaml
```diff
@@ -403,6 +403,9 @@ textures:
             winery: [792, 216, 36, 38]
             wings-joint: [756, 216, 36, 38]
             zoo: [612, 144, 36, 38]
+    grid:
```
nvkelso: Please use a more descriptive name here (like building-wall-grid or similar).

eraser-map.yaml
```diff
@@ -806,29 +809,28 @@ styles:
         base: polygons
         mix: [hsv, scale-buildings]
         texcoords: true
+        material:
+            diffuse:
+                texture: grid
```
bcamper: Need to update this to reflect texture name

bcamper:
@meetar can we to convert to PNG (since ES needs fixes for GIF)?

nvkelso:
For documentation, should all facades be defined using a PNG/GIF texture as opposed to writing a fragment shader?

bcamper:
@nvkelso which facades do you mean? All (and/or other specific ones in this style)? All in general? In this case, it is a pattern (a grid) that aliases badly when scaled, and can be easily re-created in a texture (actually the texture gets dynamically blended with the wall color, here: https://github.com/tangrams/eraser-map/pull/52/files#diff-939bb555547386cbe68c2ce87237df85R830). So it depends on the nature of the pattern and the other visual effects as to whether a texture, shader, or both is most appropriate.

bcamper:
(By using a texture we get hardware mipmapping, which automatically scales patterns like a grid well across zoom levels.)

bcamper:
Or @blair1618 is going to look at enabling GIF support for ES, which we need anyway... so.. either way :)

nvkelso:
@bcamper Thanks for the background / documentation :)

matteblair:
GIF images are working in tangram-es, just had to update stb_image to fix a bug in GIF transparency.

sensescape:
looks good, thank you : )

PR #53, opened 2015-10-22
only inject dot methods for fragment shader

tallytalwar:
Fixes: https://github.com/tangrams/tangram-es/issues/358

Should we have specific injection points for global-fragment or global-vertex?
Or maybe some other approach, to make things more intuitive and simpler for a user?

### DIFF
--- a/eraser-map.yaml
+++ b/eraser-map.yaml
@@ -676,9 +676,11 @@ styles:
                     //
                     // Get the coordinates in tile space
                     // ================================
-                    vec2 getTileCoords () {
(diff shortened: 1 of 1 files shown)


bcamper:
Discussed with @karimnaaji as well. I think for now we should stick to a simpler solution like this, and consider other ways to avoid vertex/fragment shader confusion in the future (could be separate injection points though that's more for users to understand/think about, could be as simple as avoiding using varyings in functions and sending them as function inputs instead, etc.).

PR #56, opened 2015-10-23
Use non-view space normals

### LINKED ISSUE #51: ANDROID: tilted buildings grid pattern is broken
screenshot_2015-10-08-13-51-34

When I tilt the map, the grid shader doesn't draw on the extruded buildings
cc @bcamper @patriciogonzalezvivo @nvkelso

karimnaaji:
We'd prefer to add a built-in function instead, something like `WorldSpaceNormal()` instead of accessing the attribute, but fixes #51. 
screen shot 2015-10-23 at 4 49 40 pm
cc @bcamper

### DIFF
--- a/eraser-map.yaml
+++ b/eraser-map.yaml
@@ -461,7 +461,14 @@ styles:
                     color = mix(v_color,foreground, stripes(st) );
                     //color = mix(v_color,foreground, clamp(stripes(st)-xMargin(st,.9),0.,1.) ) ;
 
-
+    tools-world-space-normal:
+        base: polygons
+        shaders:
+            blocks:
+                global: |
+                    varying vec3 v_worldSpaceNormal;
+                position: |
+                    v_worldSpaceNormal = a_normal;
 
     tools:
         base: polygons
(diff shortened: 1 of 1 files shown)


bcamper:
Makes sense. So to add a `worldNormal()` (if we're following `modelPosition()` and `worldPosition()`) we would have to add another varying, right? Kind of a bummer but I'm not sure if we have a good way around that. Sounds like it would just set a varying to the value of the attribute, return the function would return that varying, like your fix here, is that correct? 

@karimnaaji @blair1618

matteblair:
I think we could also multiply the normal varying by the normal matrix
inverse to get it back into world space, which is more expensive than using
a varying but if you don't use it then you don't pay for it.

bcamper:
I see, so I was thinking in JS it would just be `vec3 worldNormal() { return v_normal; }`... but then I realized we do actually have a normal matrix, even though it hasn't really been needed (yet): https://github.com/tangrams/tangram/blob/master/src/scene.js#L741

So... we would perhaps do the same inverse on both platforms, for consistency?

karimnaaji:
Yeah normal matrix inverse could be another uniform; actually it might be better than interpolating a default varying that most of the shaders won't use, and like @blair1618 said, you don't have to apply the inverse transform if you don't use world normals.

PR #58, opened 2015-10-26
Convert tile-based shaders to use modelPosition() accessor

bcamper:
Compatibility with JS (once Tangram 0.4.0 is released) and ES (since they pack vertex position values slightly differently right now... and these are low-level values where we want to keep the implementation opaque anyway).


bcamper:
Don't merge until Tangram JS 0.4.0 is released!

bcamper:
Superseded by #73

PR #67, opened 2015-10-30
Placeholder for route-icon

tallytalwar:


eraser-map.yaml
```diff
@@ -902,6 +902,13 @@ layers:
                 priority: 1
                 sprite: capital-xl
                 size: 18px
+    route_icon:
+        data: { source: route_icon }
+        draw:
+            icons:
+                priority: 1
+                sprite: airport
+                size: 18px
```
karimnaaji, line 911: Can you add `collide: false`, this won't make it disappear like mentioned here https://github.com/mapzen/eraser-map/issues/84#issuecomment-152560078

karimnaaji, line 911: Also it would make sense to setup transition to not fade:

``` yaml
route_icon:
    draw:
        icons:
            transition:
                [show, hide]:
                    time: 0s
```

nvkelso, line 911: Nice!

tallytalwar, line 911: Yup will do... sorry I accidentally merged this PR, will create another PR with `collide: false` and `transition` fixes.

karimnaaji:
The docs related to this are here https://github.com/tangrams/tangram-docs/pull/50.

eraser-map.yaml
tallytalwar, line 911: Here you go: https://github.com/tangrams/eraser-map/pull/68

PR #68, opened 2015-10-30
find_me, search and route_icon collision free

tallytalwar:
also with 0 transition when added/removed

### DIFF
--- a/eraser-map.yaml
+++ b/eraser-map.yaml
@@ -899,21 +899,27 @@ layers:
         data: { source: find_me }
         draw:
             icons:
-                priority: 1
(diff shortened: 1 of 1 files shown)


nvkelso:
@tallytalwar This PR looks to be outdated compared to https://github.com/tangrams/eraser-map/pull/84/files which has already been merged.

The only unique part of this PR was removing the priority, should that still happen?

matteblair:
I believe we can consider this superseded by #84

tallytalwar:
There is no use of setting priority since collide is false for these.
Priority is only used in collision resolution.

PR #73, opened 2015-11-02
Upgrade to Tangram JS 0.4

bcamper:
Small shader and JS event handling changes for Tangram JS 0.4.x compatibility.

### DIFF
--- a/eraser-map.yaml
+++ b/eraser-map.yaml
@@ -478,12 +478,12 @@ styles:
                 global: |
                     // GridTile
                     //=============================
-                    varying vec3 v_pos;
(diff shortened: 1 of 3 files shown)


bcamper:
@blair1618 this uses the accessors, so should hopefully obviate the need for the Android branch?

matteblair:
Yes, the only other mobile-related concern for the `gh-pages` branch of
eraser-map is the size of the icons texture. Since the texture size is a
known issue that we plan to address, we can bring `gh-pages` up-to-date
with `android` and start using it for the mobile app.

PR #77, opened 2015-11-04
Clean up shaders

matteblair:
Being the first in a series of pull requests in which I will be bringing the `gh-pages` branch in sync with the `android` branch.

Changes here look messy, I would suggest just inspecting the final code - the result is much more concise and readable than before
- Added conventional spacing to GLSL for readability
- Removed redundant code blocks
- Organized "tool" styles by units of functionality

### DIFF
--- a/eraser-map.yaml
+++ b/eraser-map.yaml
@@ -440,40 +440,48 @@ styles:
                     vec2 mirrorTile(vec2 _st, vec2 _zoom){
                         _st *= _zoom;
                         if (fract(_st.y * 0.5) > 0.5){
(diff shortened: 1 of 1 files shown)


matteblair:
Result can be inspected here: http://tangrams.github.io/tangram-play/?scene=https://raw.githubusercontent.com/tangrams/eraser-map/shader-cleanup/eraser-map.yaml#15.483333333333526/40.7076/-74.0095

As far as I can tell, this preserves all of our shading beauty. If nobody has concerns I'll go ahead and merge.

PR #80, opened 2015-11-09
Syntax fixes

matteblair:


matteblair:
@nvkelso is this in line with your intent for these rules?

eraser-map.yaml
```diff
@@ -1979,13 +1979,17 @@ layers:
                     color: '#444'
                     width: [[14, 0.5px], [15, 1.0px], [16, 2m]]
             gondola_cable_car:
-                lines:
-                    color: black
-                    width: [[12, 2.0px], [13, 1.75px], [14, 2.25px], [15, 3.0px], [16, 4m]]
+                filter: { kind: [gondola, cable_car] }
+                draw:
```
bcamper, line 2027: @nvkelso this and the one below need to go back to Skin & Bones

eraser-map.yaml
```diff
@@ -3940,8 +3945,8 @@ layers:
                     draw:
                         text:
                             visible: false
-                    icons:
-                        size: [[13, 8px], [19, 18px]]
+                        icons:
```
bcamper, line 4382: @nvkelso also probably needs to go to Skin & Bones

eraser-map.yaml
```diff
@@ -2204,8 +2208,9 @@ layers:
         data: { source: osm, layer: buildings }
         filter: 
             all:
-                - { $zoom: { min: 20 }, kind: address }
-                - { $zoom: { min: 20 }, [ kind: true, addr_housenumber: true, name: false ] }
+                - $zoom: { min: 20 }
```
bcamper: I'm not clear of the intent here and whether the old or new syntax achieves it :)  @nvkelso @blair1618 

My read is that this should be:

```
filter:
   $zoom: { min: 20 }
   any:
      - kind: address
      - { kind: true, addr_housenumber: true, name: false }
```

Zoom is >= 20, AND EITHER kind = address OR (kind and addr_housenumber have values, but name does not have a value).

nvkelso:
@blair1618 When I try this in Tangram Play it crashes. When I try it locally I get:

```
Unhandled promise rejection Expected () to start arrow function, but got ')' instead of '=>': SyntaxError: Expected () to start arrow function, but got ')' instead of '=>'
```

And some other debug:

```
 at Function (native)
    at Object.b [as match] (https://mapzen.com/tangram/0.4/tangram.min.js:8:7564)
    at t.m.value (https://mapzen.com/tangram/0.4/tangram.min.js:14:304)
    at t.e (https://mapzen.com/tangram/0.4/tangram.min.js:14:107)
    at new t (https://mapzen.com/tangram/0.4/tangram.min.js:14:862)
    at f (https://mapzen.com/tangram/0.4/tangram.min.js:13:30309)
    at Object.h [as parseRules] (https://mapzen.com/tangram/0.4/tangram.min.js:13:30529)
    at c.default.isWorkerThread.i.updateConfig (https://mapzen.com/tangram/0.4/tangram.min.js:12:21933)
    at DedicatedWorkerGlobalScope.<anonymous> (https://mapzen.com/tangram/0.4/tangram.min.js:16:10817)
```

bcamper:
@nvkelso @blair1618 looking into it...

bcamper:
OK, the issue appears to be a spectacular mis-parsing of this line: https://github.com/tangrams/eraser-map/pull/80/files#diff-939bb555547386cbe68c2ce87237df85R2213

Last line of:

```
    address-labels:
        data: { source: osm, layer: buildings }
        filter: 
            all:
                - $zoom: { min: 20 }
                - kind: address
                - [addr_housenumber: true, name: false]
```

This, `[addr_housenumber: true, name: false]`, is not valid in that context. The `all` filter is trying match each of the following items of the array (denoted in YAML with a `-`) as a filter. A bare array is not a valid filter object; arrays are used to match a property to multiple values (`kind: [restaurant, shop]`), or to list multiple filter objects for the `all`, `any`, or `none` keywords to match against. 

Clearly more input validation is needed on the Tangram, side here, so the filter is rejected gracefully with an error message, and the rest of the scene is parsed.

In the meantime, can we clarify the intent for when address labels are applied? Is it at zoom 20, when either `kind` is `address`, or `addr_housenumber` is not null but name _is_ null? If so, that would be most concisely expressed as:

```
filter:
   $zoom: { min: 20 }
   any:
      - kind: address
      - { addr_housenumber: true, name: false }
```

matteblair:
The doc page on filters made me believe that an array is a valid filter object ([this part](https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/Filters-Overview.md#lists-imply-any-mappings-imply-all)). Am I misreading it? Should the docs be changed?

bcamper:
Eh, that appears to be incorrect :/  I can't find any code that would have
behaved that way (though I vaguely recall discussing that idea).

What do we do in ES? We can clarify docs, and/or bring JS into line if ES
supports this.

matteblair:
I see! ES does accept this as a filter (which is why I thought it was valid, oops), it just arose naturally when we implemented filter parsing recursively. Allowing this seems consistent with the logic of filter structures.

bcamper:
I agree, implicit `any` there does make sense. I will amend JS (and add better validation in general).

bcamper:
Will update here when fixed.

tallytalwar:
Interesting, As Matt mentioned, we are handing an `array` in filter as a `any` on all elements of the array, which is precisely what is documented, and also make sense to me.

tallytalwar:
Just saw your comment, Brett. :+1:

bcamper:
To be clear, you mean that when a filter object is an array, it is interpreted as if it is an `any` object, meaning:

`[addr_housenumber: true, name: false]`

is equivalent to:

`{ any: [addr_housenumber: true, name: false] }`

Correct? I think the only reason we didn't do this (I remember discussing it) was potential confusion over the subtly of `{}` vs.`[]` having very different behavior. But I also think it makes sense (especially with implicit `all` already) and will update in JS.

tallytalwar:
Yup, when a filter object is an `array` it will be interpreted as `any` operator, and when a filter object is a `map`, it will be interpreted as `all` operator.

bcamper:
Should be resolved in: https://github.com/tangrams/tangram/commit/6a785eae0f83c3e4bee6775834179ad7c56c02a6

Will confirm once Tangram 0.4.1 is released.

matteblair:
Awesome!

In the meantime, I've revised the filter to use @bcamper 's suggestion. @nvkelso does this look good to you?

bcamper:
Tangram 0.4.1 is out, this branch loads OK:

https://tangrams.github.io/tangram-play/?scene=https://raw.githubusercontent.com/tangrams/eraser-map/syntax-fixes/eraser-map.yaml

PR #81, opened 2015-11-09
Fix landuse layers with no order defined

matteblair:
This problem isn't visually observable in tangram-js (yet), but there are many landuse layers that do not have any 'order' assigned. In tangram-es currently (and in tangram-js, in the near future), this will cause those features to not be drawn. 

This assigns an order for all existing landuse layers using the over-arching z-order rule. Setting visibility explicitly is necessary to avoid drawing features that are matched by the z-order rule, but not by any sub-rules.

### DIFF
--- a/eraser-map.yaml
+++ b/eraser-map.yaml
@@ -4482,13 +4482,14 @@ layers:
                     dots2:
                         color: [[4,[0.655,0.800,0.796]],[8,[0.588,0.780,0.773]]]
                         order: 0
-                        # visible: false
+                        visible: true
                 us_national_park:
                     # yosemite national park, death valley national park, grand canyon national park
                     filter: { operator: [ "United States National Park Service", "US National Park Service" ] }
                     draw:
                         dots2:
                             color: [[4,[0.000,0.745,0.497]],[9,[0.000,0.631,0.421]]]
+                            visible: true
                             # color: [[4,[0.242,0.737,0.712]],[12,[0.224,0.682,0.659]]]
         tier2:
             filter: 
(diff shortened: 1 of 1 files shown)


nvkelso:
Why are we defining all these to be false to then set the visibility to all be true? Should the style be refactored (probably), or is there a Tangram bug we're working around or?

matteblair:
Not a bug in tangram, just a consequence of the style rule behavior. Here's the situation:

To specify an `order` for the landuse layers that use the "dots2" style, we can either add an `order` rule to every draw group in those layers _or_ we can follow the pattern used elsewhere of matching many layers at once with a "z-order" layer that has no filter. When we do that (like [this](https://github.com/tangrams/eraser-map/pull/81/files#diff-939bb555547386cbe68c2ce87237df85R4796)) we suddenly see lots of areas of dots being drawn with a color that isn't in the stylesheet at all - how curious! What's going on is that every feature in the "landuse" layer is matching a draw rule with `style: dots2` and `order: function () ...`, the rule we just created for "z-order". To be considered valid for drawing, a feature must match a rule that has an `order` and either a `color` _or_ a style that has a "color" or "filter" block in its shader. The "dots2" style has a "color" shader block, so all of those landuse features are perfectly valid for drawing, even though the only rule they matched was for "z-order"! These features will therefore all be drawn in the "dots2" style and their color will be essentially undefined (happens to be yellow in tangram-es). 

Adding these `visible` rules was the simplest way I could find to create the desired behavior :\

matteblair:
Since this behavior isn't a problem for styles without custom shaders, an alternative fix would be to set the style explicitly to "dots2" in all the landuse layers and set the style explicitly to "polygons" in the "z-order" layer. Doesn't seem much more elegant to me.

nvkelso:
Wow, that's complicated. Thanks for the detailed explanation! Agree that we should just merge this and table the architecture bits till later.

PR #82, opened 2015-11-10
Functions accessing varyings can only be defined in the fragment shader

matteblair:
Fixes https://github.com/tangrams/eraser-map/issues/69 and similar issues on other phones.

### DIFF
--- a/eraser-map.yaml
+++ b/eraser-map.yaml
@@ -491,12 +491,11 @@ styles:
                 global: |
                     // Varying to be added to both vertex and fragment shaders
                     varying vec2 v_pos;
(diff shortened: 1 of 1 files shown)


tallytalwar:
Is this defined in glsl spec?

matteblair:
@tallytalwar not that I've found so far. It might be part of the spec, but this is just based on empirical evidence with the drivers we've been able to test.

tallytalwar:
Yup. We gave had a couple of solid evidences showing this behavior.

nvkelso:
Rock it.

PR #84, opened 2015-11-10
Disable collisions and transitions for dynamic data

matteblair:
`transition` is an ES-only property for now, but turning it off for the route icon and location gem keeps them from looking like pacman ghosts when they're updated frequently :)


nvkelso:
LGTM

tallytalwar:
You could have just cherry picked the commit from android branch. :)

PR #89, opened 2015-11-16
Anchor: top on search and route icon

tallytalwar:
also no priority required for collide:false icons

### DIFF
--- a/eraser-map.yaml
+++ b/eraser-map.yaml
@@ -831,7 +831,7 @@ layers:
         data: { source: route_icon }
         draw:
             icons:
-                sprite: route-start
+                sprite: current-location
                 size: [38px,49px]
                 collide: false
                 transition:
--- a/eraser-map.yaml
+++ b/eraser-map.yaml
@@ -821,7 +821,6 @@ layers:
         data: { source: find_me }
         draw:
             icons:
-                priority: 1
                 sprite: current-location
                 size: 36px
                 collide: false
@@ -832,10 +831,10 @@ layers:
         data: { source: route_icon }
         draw:
             icons:
-                priority: 1
                 sprite: route-start
                 size: [38px,49px]
                 collide: false
+                anchor: top
                 transition:
                     [show, hide]:
                         time: 0s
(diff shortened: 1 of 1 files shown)


eraser-map.yaml
```diff
@@ -832,10 +831,10 @@ layers:
         data: { source: route_icon }
         draw:
             icons:
-                priority: 1
                 sprite: route-start
                 size: [38px,49px]
                 collide: false
+                anchor: top
```
nvkelso: Why isn't this center?

eraser-map.yaml
```diff
@@ -843,10 +842,10 @@ layers:
         data: { source: search }
         draw:
             icons:
-                priority: 2
                 sprite: search-active
                 size: [40px,55px]
                 collide: false
+                anchor: top
```
nvkelso, line 847: Why isn't this bottom?

eraser-map.yaml
```diff
@@ -821,7 +821,6 @@ layers:
         data: { source: find_me }
         draw:
             icons:
-                priority: 1
```
nvkelso, line 824: Should this also have an anchor (center)?

tallytalwar, line 824: I think center is the default. Yup center makes sense for this.

eraser-map.yaml
tallytalwar: I tried `top` and `bottom` and `bottom` looked better to me, considering we have a bubble icon. 

But its a designer's choice, so I would leave the call for setting of anchors to you and Geraldine :D.

nvkelso: Talking more with Chuck about this, we should be using a different artwork asset for this (more like a arrow in a circle badge), and center (default anchor) would make more sense for this.

eraser-map.yaml
nvkelso, line 847: LGTM.

nvkelso, line 847: (anchor is with respect to the feature XY, not the icon bounds)

PR #93, opened 2015-11-23
Nvkelso/overlay ux pieces

nvkelso:
- move blend:overlay to new style classes, and use those in app UX blocks
- rebase against gh-pages branch


nvkelso:
Brett gave thumbs up, Geraldine and I are merging.

eraser-map.yaml
```diff
@@ -794,6 +794,16 @@ styles:
         base: points
         texture: pois
         interactive: true
+        
+    icons-overlay:
+        base: points
+        texture: pois
+        interactive: true
+        blend: overlay
```
matteblair, line 802: You don't actually need `icons-overlay` because the default blend mode for the points style is overlay. Was the intent to make these icons draw over labels?

nvkelso, line 802: Yes, these icons (for routing and search) should always appear **above** other map icons. Do I set that with an order or priority?

matteblair, line 802: At present, I don't think there's any way to guarantee this in tangram-es. 

`priority` only affects the result of label collisions, and collision is turned off for routing and search results. 

`order` would conceptually fit this need, but currently it has no effect on `overlay` styles (features in these styles are all drawn at the same depth).

eraser-map.yaml
```diff
@@ -794,6 +794,16 @@ styles:
         base: points
         texture: pois
         interactive: true
+        
+    icons-overlay:
+        base: points
+        texture: pois
+        interactive: true
```
matteblair, line 801: In this style and the one above it, `interactive` is not a valid parameter. `interactive` is a parameter for draw rules, on styles it will do nothing. (see the [docs](https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/draw.md#interactive))

eraser-map.yaml
nvkelso, line 802: Geraldine modified this in: 
https://github.com/tangrams/eraser-map/commit/8ee756535027d24b056242429a42da34c27022e0

To remove the `icon-overlay` style and reference, and move the `lines-overlay` above the `icons` style so that icons were always on top of the route line. Which doesn't make a whole lot of sense, but works.

matteblair, line 802: It looks like the `icons-overlay` style is still in there in `gh-pages`; it's harmless, but would be good to clean out at some point, along with the `interactive` parameters mentioned above. 

I'm not totally clear what you mean by moving `lines-overlay` "above" `icons`, but I'll mention that the ordering of styles in the scene file doesn't guarantee drawing order.

PR #105, opened 2015-12-01
Fix: undefined behavior for region rule match

tallytalwar:
Similar to https://github.com/tangrams/eraser-map/issues/101, this fixes an issue with parallel rule matching for region
names.

NOTE: I am not sure what should be styling behavior here with respect to design point of you, so please feel free to
change it to reflect better design.

cc. @nvkelso


nvkelso:
LGTM

PR #111, opened 2015-12-04
Add bundled font for distribution with app

matteblair:
Adds all 10 available variants of OpenSans as TTF files (plus Apache license)

Replaces Helvetica with OpenSans in stylesheet; OpenSans isn't available in weight 500, so weight was adjusted to 600 in some places

The font may not load in tangram-js if OpenSans isn't installed on your machine (you can download the font [here](http://www.opensans.com/download/open-sans.zip) and install it).


tallytalwar:
Matt we might need all lowercase for file name.

matteblair:
Hmm yeah I remember now that we decided to require all-lowercase names, but it also works with the current capitalization. I'll check it out!

tallytalwar:
:+1

tallytalwar:
Another thing I had modified in my local testing was weights like `bold`, etc.

So besides numeric weights there can also be `bold`, `light`, etc... there should be ttf files available for these also. Can you check this also?

matteblair:
1. Yes it probably matters that the font file names be all-lowercase; I had forgotten that the current file system on OS X is actually case-insensitive. The file names are all-lowercase now. 
2. The fonts were initially labeled with english weights ("light", "bold", etc.) but since numeric weights were more common in the stylesheet I changed the file names to use the numeric equivalents (there's a one-to-one mapping between the english and numeric weights, see [this](http://www.google.com/fonts/specimen/Open+Sans)).

tallytalwar:
Looks good to me.

nvkelso:
@blair1618 Looks like the bulk of the difference is I created an anchor and used that in place of the inline Helvetica and you replaced each instance of Helvetica to OpenSans inline?

Why is it `OpenSans` instead of `Open Sans`? The version with the space works on both Geraldine and my MacBooks, will the no space version?

matteblair:
Rebasing this now, I'll use the anchor as you did

The font files were named "OpenSans" when I downloaded them, that's the only reason I omitted the space. With the font installed on my machine, both "OpenSans" and "Open Sans" yielded the correct font when I tested in tangram-play.

tallytalwar:
Will "open sans" work on device? I mean the name with the space?

matteblair:
Font file names use "open sans" now and the new font seems to work on tangram-es, desktop at least. Looking pretty nice!

nvkelso:
Geraldine and I are really digging it, too! :)

nvkelso:
Looks ready to merge now?

matteblair:
Yes, just making sure it worked in the app. Should be good to merge here. I'll make one small change in tangram-es, update the style submodule in the app repo, then we should see the new font in the app.

PR #115, opened 2015-12-10
Adjust text anchors & offsets so labels are below icons

bcamper:
Uses `anchor: bottom` and the new zoom interpolated `offset` to properly position label text below their icons/POIs.

This is a bit messy to do because the text and icon are unaware of each other, and while we can set `anchor: bottom` on the text, that only anchors it to the bottom of the actual lat/lng point for the POI, it doesn't account for the actual icon height. Meanwhile,  the icon size itself is changing across zooms! :)

So the current solution was to use an interpolating `offset` that tracks to the size of the icon, so the text is always positioned down an additional half-icon-height. The longer term solution is a unified label style that can do layout for text and icon together.

With correct text placement:

screen shot 2015-12-10 at 4 10 15 pm

screen shot 2015-12-10 at 4 10 46 pm

screen shot 2015-12-10 at 4 12 29 pm

### DIFF
--- a/eraser-map.yaml
+++ b/eraser-map.yaml
@@ -2473,7 +2473,6 @@ layers:
         draw:
             text:
                 visible: false    # labels are enabled by each layer below
-                anchor: bottom
                 font:
                     family: *text_font_family
                     # weight: 500
@@ -2690,6 +2689,7 @@ layers:
                 icons:
                     priority: 5
                 text:
+                    anchor: bottom
                     priority: 6
 
             populated-places-natural-earth-z2:
@@ -2698,7 +2698,7 @@ layers:
                     text:
                         interactive: true
                         visible: *text_visible_populated_places
-                        offset: [0, 3.5px] # half icon size
+                        offset: [0, 1px] # half icon size
                         font:
                             size: 10px
                             fill: *text_fill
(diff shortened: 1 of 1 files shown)


bcamper:
Note, this `offset` syntax works with the new 0.4.4 JS build, and will work on ES as soon as this PR is merged: https://github.com/tangrams/tangram-es/pull/429, but is safe to merge in the meantime (the new syntax will just be ignored, leaving the current incorrect offset).

nvkelso:
The populated places offsets need to be tightened to reduce gap, I'll land a fix in a few minutes.

bcamper:
It's possible that is caused by JS moving the point location when near a
tile edge too (though also possible I just got the offset math wrong :)

nvkelso:
It's probably that the assets have too much transparency color around them, too? At any rate, tighter is better with circles and labels.

bcamper:
Ah yeah, I was gonna say, I was having problems getting some of these
combos to show up on ES because of overlap (which actually led to a
separate ES bug Karim is fixing). But with current ES behavior (assuming
bug is fixed), any overlap between the two will cause one to get culled
from collision. So we may need to bring down the transparent buffer in the
icon size to get them to overlap properly, as you say (and/or add an option
to allow overlap in collisions, but probably not the best solution here and
I'm not sure if it's a generally desirable feature).

nvkelso:
Let's treat refactoring the townspot artwork assets as a separate issue. I've cleaned this up so it looks good on WebGL and any ES optimizations

PR #121, opened 2015-12-16
Assign blend_order to styles

tallytalwar:
Styles should be drawn as follows:
1. ground, landuse, roads, buildings, etc.
2. route line
3. labels + icons
4. search results + route icon (app chrome)

where `n+1` is drawn above `n`.

Tested this in eraser-map app.

### DIFF
--- a/eraser-map.yaml
+++ b/eraser-map.yaml
@@ -4291,7 +4291,7 @@ layers:
                         priority: 17
 
             # add generic icon at high zoom
-                        default-label-text-blend-order:
+            default-label-names:
                 filter: { $zoom: { min: 13 }, area: false }
                 draw:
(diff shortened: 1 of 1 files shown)


tallytalwar:
cc. @nvkelso, Can you have a look at this and merge if this looks good.

This will make sure, route line is drawn below labels and icons, and search results and route icons (icon, start, stop) are drawn above everything else.

tallytalwar:
screenshot_2015-12-16-16-44-43

nvkelso:
Having a look...

nvkelso:
Working for me in app (master-587) for routing. Search is crashing the app so can't confirm there. Time for a drive test :)

PR #122, opened 2015-12-17
Improve dots shader

hjanetzek:
This should repeat the pattern correctly when rendering proxy tiles


tallytalwar:
I think @blair1618 had something similar in his testing.

nvkelso:
@blair1618 Can you review this?

matteblair:
Will do

matteblair:
Looks good, this corrects the overgrown dots

PR #123, opened 2015-12-18
Fix conflicting orders for water features

matteblair:
- Fixes order conflict between water areas and piers (https://github.com/tangrams/eraser-map/issues/114) by moving 'water' from order 5 to 2
- Fixes water line features being drawn twice with the same order (they seemed to be intended to function as an outline, so I just converted them to outlines)
- Fixes rivers drawn with the same order as some landuse areas (moved rivers from order 3 to 18)

### DIFF
--- a/eraser-map.yaml
+++ b/eraser-map.yaml
@@ -962,8 +962,8 @@ layers:
                 line2:
                     style: lines
                     order: 7
-                    width: [[9,0.25px],[10,2px],[11,3px],[13,4px],[15,6px]]
                     color: [[9,[0.396,0.833,0.920]],[10,[0.886,0.937,0.988]]]
+                    width: [[9,0.25px],[10,2px],[11,3px],[13,4px],[15,6px]]
                 lines:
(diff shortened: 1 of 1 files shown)


eraser-map.yaml
```diff
@@ -919,7 +919,7 @@ layers:
         data: { source: osm, layer: water }
         draw:
             grid:
-                order: 5
+                order: 2
```
nvkelso: Water must be on top of other landuse (else ponds and lakes inside of landuse areas disappear).

This needs to stay at `5` (but piers could move to say, `6`).

matteblair: I see! Looks like several tunnel rules have `order: 6`, but I would expect overlaps between tunnels and piers to be much less frequent than pier/water overlaps. I'll set water back to `order: 5` and piers to `order: 6`.

nvkelso: Good catch! Another reason to add more elbow room between the landuse and roads orders :)

PR #125, opened 2015-12-23
add parameter for building scale-up

### LINKED ISSUE #124: Buildings should extrude into view (not pop)
We x-fade the building extrusion so it's less extreme at street-level zooms (which is very nice). But now the big buildings just POP into full extrusion at the mid-zooms. Instead it would be nice if they "grew" into their full height (very quickly in the same integer zoom) before slowly getting shorter.

meetar:
Rename the `speed` parameter to `outspeed`, and add a new parameter `inspeed` – this changes the speed at which buildings scale up as you zoom in.

`inspeed` is currently set to .1, or one-tenth of a zoom level. Adjust to taste.

Fixes #124.

### DIFF
--- a/eraser-map.yaml
+++ b/eraser-map.yaml
@@ -758,10 +758,17 @@ styles:
                 position: |
                     // scale buildings based on zoom
                     float zoom = u_map_position.z;
-                    float min = .1;         // minimum building scale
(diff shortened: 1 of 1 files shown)


nvkelso:
Nice!

PR #132, opened 2016-01-05
Reduce outlines with minimal visual impact

bcamper:
I've found that there are a lot of reaaaally thin outlines that had minimal (practically imperceptible) visual contribution to the scene. particularly at some lower zooms. By slightly cutting down on these (introducing them a bit later), we can get some substantial geometry count / performance benefits. This PR also includes a few syntax clean-ups done while reviewing outlines, such as removing redundant zoom stops (e.g. `[9, 0.0px], [11, 0px], [12, 0px], [14, 0px]` becomes just `[14, 0px]`). Examples below. These are fairly conservative, with a more intentional design eye we could probably cut them down further.

**z8**: 277k -> 209k polys (**-68k polys**)
z8

**z9**: 435k -> 399k polys (**-36k polys**)
z9

**z10**: 285k -> 267k polys (**-18k polys**)
z10

**z11**: 194k -> 182k polys (**-12k polys**)
z11

**z12**: 251k -> 247k polys (**-4k polys**)
z12


eraser-map.yaml
```diff
@@ -1140,8 +1138,8 @@ layers:
                     color: [[8, *highway1], [14, *highway1], [15, [1.0,1.0,1.0]], [17, [1.0,1.0,1.0]]]
                     width: [[8, 1px], [9, 1.25px], [11, 1.2px], [12, 1.3px], [13, 1.5px], [14, 2.25px], [15, 0.15px], [16, 5.5px], [17, 12m]]
                     outline:
-                        color: [[9, *highway_casing1], [13, *highway_casing1], [15,[0.988,0.495,0.439]], [17, [0.988,0.495,0.439]]]
-                        width: [[9, 0.5px], [12, 1px], [15, 2px], [16, 2.25px], [17, 3px], [18, 3px]]
+                        color: [[13, *highway_casing1], [15,[0.988,0.495,0.439]], [17, [0.988,0.495,0.439]]]
+                        width: [[12, 0], [13, 1.33px], [15, 2px], [16, 2.25px], [17, 3px], [18, 3px]]
```
nvkelso: I notice the missing casing at z12 (but not at earlier zooms), and this is the start of the transition. It emphasizes large highways over major roads, and puts the highways over ramps to reduce visual clutter.

Suggest instead:

```
                        color: [[11, *highway_casing1], [15,[0.988,0.495,0.439]], [17, [0.988,0.495,0.439]]]
                        width: [[11, 0], [12, 1px], [13, 1.33px], [15, 2px], [16, 2.25px], [17, 3px], [18, 3px]]
```

eraser-map.yaml
```diff
@@ -1325,7 +1321,7 @@ layers:
                             width: [[9, 1.0px], [10, 1.0px], [11, 1.2px], [12, 1.25px], [13, 1.5px], [14, 1.6px], [15, 0.15px], [16, 5.25px], [17, 12m]]
                             outline:
                                 color: [[13, [0.870,0.870,0.870]], [15, [0.976,0.612,0.573]]]
-                                width: [[9, 0.0px], [12, 0px], [15, 2px], [16, 2px], [17, 3px]]
+                                width: [[13, 0], [15, 2px], [16, 2px], [17, 3px]]
```
nvkelso, line 1324: We use px everywhere else but here:

`[[13, 0px],`

eraser-map.yaml
```diff
@@ -1483,7 +1479,7 @@ layers:
                         width: [[11, 0.65px], [12, 0.75px], [13, 1.25px], [14, 2px], [15, 5px], [16, 10m], [17, 10m]]
                         outline:
                             color: [[15, [1.000,0.739,0.710]]]
-                            width: [[11, 0px], [14, 0px], [15, 1.75px], [16, 2px], [17, 3px]]
+                            width: [[14, 0], [15, 1.75px], [16, 2px], [17, 3px]]
```
nvkelso, line 1482: We use px everywhere else but here:

[[14, 0px],

eraser-map.yaml
```diff
@@ -1466,7 +1462,7 @@ layers:
                         width: [[9, 0px], [10, 0.5px], [14, 2px], [16, 4m], [17, 5m], [18, 6m]]
                         outline:
                             color: [[9, [0.827,0.827,0.827]], [13, [0.827,0.827,0.827]], [15, [1.000,0.739,0.710]]]
-                            width: [[9, 0px], [10, 1px], [14, 1px], [16, 1.5px], [17, 2px]]
+                            width: [[13, 1px], [14, 1px], [16, 1.5px], [17, 2px]]
```
nvkelso: At z10, what with the outline width be? 0.75px?

eraser-map.yaml
```diff
@@ -1624,7 +1620,7 @@ layers:
                         outline:
                             color: [0.827,0.827,0.827]
                             # todo: look at zoom 13 width
-                            width: [[12, 0px], [13, 0px], [14, 0px], [15, 0px], [16, 1.0px], [18, 3.5px]]
+                            width: [[15, 0], [16, 1.0px], [18, 3.5px]]
```
nvkelso, line 1623: We use px everywhere else but here:

[[15, 0px],

eraser-map.yaml
```diff
@@ -1633,7 +1629,7 @@ layers:
                             width: [[11, 0.5px], [12, 0.5px], [13, 0.65px], [14, 1.0px], [15, 2px], [16, 8m], [19, 10m]]
                             outline:
                                 color: [0.827,0.827,0.827]
-                                width: [[12, 0px], [13, 0px], [14, 0px], [15, 0px], [16, 1.0px], [18, 3.5px]]
+                                width: [[15, 0], [16, 1.0px], [18, 3.5px]]
```
nvkelso, line 1632: We use px everywhere else but here:

[[15, 0px],

eraser-map.yaml
nvkelso: But at z9 it would be 0.7px. So this is changing the look (was no width). Suggest keeping as is unless there is a strong perf reason.

bcamper: In theory it is, in the screenshots, I'm having a hard time identifying these changes except as a handful of pixels (maybe I have more tolerance). The performance impact of theses changes is real, -12% at z9 but still -7% at z11. So if you see specific issues I think we can likely adjust the style around it, but as is it just did not seem worth adding the weight to the scene for what practically amounts to noise IMO :)  

Sub-pixel outlines can also cause aliasing artifacts so I think it's also worth trying out alternatives, in some cases it may look better without (this is subjective though). I particularly want to avoid sub-pixel outlines at z8-10 where possible. Even when the width is 0px at a particular zoom, the line is built and rendered at that zoom if the _next_ zoom has a width, because it's interpolating to the next zoom's value.

If we need to go earlier here perhaps we could start the outline at z12 instead of z13 (which means it would build lines at z11), or is there a bigger visual impact I'm not seeing?

eraser-map.yaml
bcamper: I think that's fine since highways at z12 are not so high-density compared to outline rules affecting all roads.

eraser-map.yaml
bcamper, line 1324: Yeah I was sloppy about this, I originally was going to update all the `0px` to be `0`, because there should be an advantage to not having to do unit conversion from pixels to meters, but then didn't want to expand scope of the PR. This may be silly though, Tangram itself can just remove the units for a zero value. (There's still probably a slight parsing advantage because it only has to look for units if the value is a JS string, if it's a pure JS number it already knows there's not unit.) I can benchmark to see how meaningful (if at all) the difference is.

eraser-map.yaml
nvkelso: I notice the change at z11 and z12, and it's a design detail that Geraldine and I were deliberate about. I'd like to keep the casing at these two zooms. I'm fine removing it at z9 and z10, and starting z11 with full 1px size.

nvkelso: (These are like text halos for roads, they add to the visual hierarchy in subtle ways that seem minor in isolation, but in aggregate across the whole screen make a difference.)

eraser-map.yaml
nvkelso, line 1324: I think the PR can land with either format (0 or 0px), as long as you link your larger Tangram issue here ;)

bcamper:
Thanks!

PR #136, opened 2016-01-14
Interactive is a draw property and not style prop

tallytalwar:
- Fix search icons to be interactive

Previously `interactive` was set as a style property for `icons` and `icons-overlay` style. But it should be the `draw` group property. I have added `interactive: true` explicitly for client data icons (example: search, current-location, etc). 

I do not know if the intend is to set `interactive: true` for all icons, if so can update all `draw` groups drawing `icons` style to have this property.

cc. @nvkelso

### DIFF
--- a/eraser-map.yaml
+++ b/eraser-map.yaml
@@ -861,7 +861,6 @@ styles:
     icons:
         base: points
         texture: pois
-        interactive: true
         blend_order: 1
     text-blend-order:
         base: text
@@ -869,7 +868,6 @@ styles:
     icons-overlay:
         base: points
         texture: pois
-        interactive: true
         blend: overlay
         blend_order: 2
     lines-overlay:
@@ -894,6 +892,7 @@ layers:
         data: { source: find_me }
         draw:
             icons-overlay:
+                interactive: true
                 sprite: current-location
                 size: 36px
                 collide: false
@@ -904,6 +903,7 @@ layers:
         data: { source: route_icon }
         draw:
             icons-overlay:
+                interactive: true
                 sprite: route-arrow
                 size: [33px,40px]
                 collide: false
@@ -914,6 +914,7 @@ layers:
         data: { source: route_start }
         draw:
             icons-overlay:
+                interactive: true
                 priority: 1
                 sprite: route-start
                 size: [36px,46px]
@@ -923,6 +924,7 @@ layers:
         data: { source: route_stop }
         draw:
             icons-overlay:
+                interactive: true
                 priority: 1
                 sprite: route-stop
                 size: [36px,46px]
@@ -932,6 +934,7 @@ layers:
         data: { source: search }
         draw:
             icons-overlay:
+                interactive: true
                 sprite: search-active
                 size: [36px,46px]
                 collide: false


eraser-map.yaml
```diff
@@ -861,7 +861,6 @@ styles:
     icons:
         base: points
         texture: pois
-        interactive: true
```
nvkelso, line 864: @bcamper Not for this PR, but can this be made to work? It's confusing why this doesn't inherit.

bcamper, line 864: The parameters in the `draw` blocks are all parameters that can vary per-feature, while those for a `style` affect everything rendered with that style. Most of the time they are dealing with pretty different things, and currently I don't believe there's any overlap between them. Sometimes the distinction is reflective of underlying technical architecture and doesn't feel natural, which I'd like to address (an example is having to create a `style` for each unique `texture`, so that the texture is bound in GL when the style is rendered, but then the specific `sprite` that is used within that texture is determined per-feature in the `draw` group). The reason `interactive` is a `draw` parameter is because it's often desirable to set the interactivity based on feature criteria, rather than enabling it in a blanket fashion. That said, I can see how it might be useful to use some `style`-level parameters as default values that can be overridden in the `draw` groups. I'll make a note for future architecture/syntax discussion.

bcamper, line 864: In case this clarifies at all (maybe not :), everything in `style` affects _GL render state_, nothing technically "inherits" to the draw group. Whereas draw groups nested in sub-layers inherit properties from those above them. So properly speaking if we made a connection here it would probably make more sense to think of it as a default value than an inherited one.

nvkelso, line 864: What about something like...

```
icons:
    base: points
    texture: pois
    draw:
        interactive: true
```

tallytalwar:
@nvkelso Based on this change other interactivity might have to be updated for all other icons in the style sheet. Also should labels have interactivity?

nvkelso:
@tallytalwar Can you confirm that the old `interactive: true` declaration on line 864 isn't doing anything? So removing it won't cause any regressions. Later on in the stylesheet most the things that _should_ be interactive are additionally declared interactive already.

tallytalwar:
@nvkelso Yup, https://github.com/tangrams/eraser-map/pull/136/files#diff-939bb555547386cbe68c2ce87237df85L864 has no affect on interactivity. As @bcamper mentioned a style property does not get trickled down to draw rules.

eraser-map.yaml
bcamper, line 864: Yep, that was what I was thinking. I'd want to first look at how common
this actually is though, it fits into a more general semantic division
between styles and draw groups, which we may be able to improve in other
ways.

tallytalwar:
@nvkelso Another thing which can be done here is to exploit draw rules inheritance! So, set the uber poi layer to have `interactive: true`, and only override it to `interactive: false` for whatever filter rules/features you do not want to have interaction:

Example: https://github.com/tangrams/eraser-map/commit/ddf5da177ea855ec162b40e26c16ac53c69901a2 (removed 114 lines :D)

nvkelso:
LGTM

PR #139, opened 2016-01-14
interactive using draw rule inheritance

tallytalwar:
Use inheritance rules to grab `interactive: true` from parent layer.


nvkelso:
LGTM

PR #145, opened 2016-01-27
Remove transition to some label icons

karimnaaji:
Don't apply fading transition to
- search
- reverse_geocode
- route_start
- route_stop

cc @nvkelso


nvkelso:
LGTM

PR #165, opened 2016-03-03
Transform icon sheet to prevent artifacts from transparency blending

matteblair:
Transparent pixels now have RGB values calculated as the average of their non-transparent neighbor pixels.


nvkelso:
The spritesheet is generated out of Adobe Illustrator. Next time we update it, what process do I need to run to transform the RGB values like you've done in this commit?

nvkelso:
Otherwise looks good.

tallytalwar:
I guess you will have to run the python script @blair1618 posted on slack.

matteblair:
@nvkelso The script is here, with usage comments: 

https://gist.github.com/blair1618/63d7d5459e3edc56aafe

We could track it in git if you prefer.

nvkelso:
Tracking it in the gist for now is fine. Really a sprite artwork repo with the Illustrator file and export JSX file should be created, and the python file would be hosted there, I think?

PR #167, opened 2016-03-07
nvkelso/v0.8 tiles sort key orders

nvkelso:
**¡¡¡CAUTION!!!** Don't merge this until v0.8 tiles are live on prod!
- remove or update hard coded order values in scene file in favor of server values
- other polish


nvkelso:
v0.8 tiles are live in prod now.

PR #174, opened 2016-03-29
Load Open Sans as web font

bcamper:
The Bubble Wrap style is designed to use Open Sans as the font face. Tangram ES uses the font files bundled in this repo, but Tangram JS currently only renders with custom fonts if they are installed on the user's system. This PR loads Open Sans as a web font (via Google Fonts), then uses the [`FontFaceObserver`](https://github.com/bramstein/fontfaceobserver) library to wait until the font loads before rendering Tangram (this is needed because browsers do not guarantee the font is loaded before it is used for rendering, and typically it is not loaded until the first tile or two has rendered).

Tangram will [support this directly](https://github.com/tangrams/tangram/issues/281) in a future version, but there are still a few issues to work out. In the meantime, this demonstrates the concept and provides an example for others who may want to use custom fonts in their styles. (Note: this PR could load the TrueType files bundled in the repo, but for web use, the Google Fonts are going to be smaller and faster, and were simpler to add).


bcamper:
cc @meetar @burritojustice here's an example of how this works, will be into Tangram but can use this approach in the meantime.

nvkelso:
Neat!

PR #176, opened 2016-04-15
Standardize and namespace styles

sarahcodes100:


ecgreb:
One thing I'm still not sure about is do we need to update the layer name and source name or just the latter?

bubble-wrap.yaml
```diff
@@ -787,16 +787,16 @@ scene:
         color: *scene1
 
 layers:
-    # Map overlays for styling the server response (using special source layer names) for route line, find_me location gem, and search result pins
-    route:
-        data: { source: route }
+    # Map overlays for styling the server response (using special source layer names) for mz_route_line, mz_current_location location gem, and mz_search_result result pins
```
nvkelso: These should be plain english versions of the features, not variable names. 

Update to:
- `for route line, current location, and search result pins`

bubble-wrap.yaml
```diff
@@ -787,16 +787,16 @@ scene:
...
+    mz_route_line:
+        data: { source: mz_route_line }
         draw:
             ui-lines-overlay:
                 color: '#06a6d4'
                 order: 500
                 width: [[0,3.5px],[5,5px],[9,7px],[10,6px],[11,6px],[13,8px],[14,9px],[15,10px],[16,11px],[17,12px],[18,10px]]
     current_location_gem:
```
nvkelso: mz_current_location:

nvkelso:
I'm in favor of updating both (as Sarah has proposed).

sarahcodes100:
It looks like we do need them both updated

nvkelso:
LGTM

PR #178, opened 2016-04-18
add default point, line, and polygon overlays for SDK

nvkelso:
- add new layers & styling for SDK point, line, polygon overlays
- add new primitive "styles" for point, line, polygon overlays
- add new example GeoJSON sources for point, line, polygon default overlays

### DIFF
--- a/bubble-wrap.yaml
+++ b/bubble-wrap.yaml
@@ -418,17 +418,17 @@ sources:
     #     type: GeoJSON
     #     url: https://gist.githubusercontent.com/anonymous/57dc09eeb120919f76de/raw/43426217da3c2bae0522dc4257aaa61e4df3981e/map.geojson
     # Default point styling (SDK)
-    mz_default_point:
-        type: GeoJSON
-        url: https://gist.githubusercontent.com/anonymous/16324c771edfce45be0721390389b878/raw/7dbaebf17da7da8562e6c6f8768bc8cff83efa88/map.geojson
-    # Default line styling (SDK)
-    mz_default_line:
-        type: GeoJSON
-        url: https://gist.githubusercontent.com/anonymous/26f4e8b6b34b2617b5d5533d89decb39/raw/df8e180ab4f7f19448014dccc4a755f7cfa20003/map.geojson
-    # Default polygon styling (SDK)
-    mz_default_polygon:
-        type: GeoJSON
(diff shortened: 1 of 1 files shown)


nvkelso:
@sarahlensing @bcamper for review.

nvkelso:
Like the UX stuff, we'll want this in each Mapzen house cartography style, but will modify the colors, line widths, etc to match the appearance of each house style.

nvkelso:
Preview:

screen shot 2016-04-18 at 11 18 40

bubble-wrap.yaml
```diff
@@ -852,6 +878,37 @@ layers:
...
+                    [show, hide]:
+                        time: 0s
+    mz_default_line:
+        data: { source: mz_default_line }
+        draw:
+            sdk-line-overlay:
+                color: '#06a6d4'
+                order: 503
+                width: 3px
+    mz_default_polygon:
```
nvkelso, line 902: @bcamper Thinking about John's use case for underlays, we should probably support that as well officially via a method like this?

bcamper, line 902: Do you mean have separate polygon options for underlay or overlay? Or have the SDK set different order values on the fly (probably possible)?

nvkelso, line 902: `order: 501` is a magic number, and the "underlay" version would similarly be a magic number. 

For the SDK: Would the new v0.7 global syntax help us here – so the SDK could modify the order to the "underlay" value, without needing to know it's specific value? 

For people like John: seems like we should have a "underlay" default and "overall" style defaults for them to use that would have the right properties out of the box?

bubble-wrap.yaml
```diff
@@ -852,6 +878,37 @@ layers:
...
+        data: { source: mz_default_polygon }
+        draw:
+            sdk-polygon-overlay:
+                color: [0.02,0.65,0.82,0.5]  #'#06b1e2'
+                order: 501
+            sdk-line-overlay:
+                color: '#06a6d4'
+                order: 502
+                width: 0px
```
nvkelso, line 911: I'm setting this to `0px` here to disable the stroke on the polygon, but if someone wanted to draw stroked polygons the'd set the width and it'd show up blue line by default. But it's not clear to me how the SDK would modify these values / if this is the best approach to take.

bcamper, line 911: It's a little wasteful but it works, and only affects data being fed to these styles (presumably very little), so seems fine for now.

PR #182, opened 2016-05-12
Fix tinting on building walls

bcamper:
I found another casualty related to the shader block changes from https://github.com/tangrams/tangram/pull/271 and related PRs (which affected both Tangram JS and ES).

In brief, we stopped executing the `color` shader block in the vertex shader in all cases (partially to enable mixing standard lighting with custom `color` in the fragment shader, such as for procedural patterns).

However, Bubble Wrap was relying on `color` to execute in the vertex shader, to optimize calculations that we know are constant across the surface. These stopped working, causing the buildings to lose their darkened tint on their walls.

These changes were also made in Tangram ES. This behavior is not observed in Eraser Map currently because it's built against an older version of ES, but I verified that the same issue is present in ES master.

This PR fixes the issue by moving the calculations back to the fragment shader. This isn't ideal either since it does the HSV calculations per pixel, when they are only needed per vertex (actually per face!). But it fixes the problem for now. (Caveat, I don't know if the _new_ scene file here works with the _old_ version of ES.)

We could also consider an additional vertex shader block or another configuration, to assist in optimizing cases like this. (It is technically possible to achieve a similar optimization with current behavior by hijacking the `normal` block, but this seems like a bad idea because it makes the same kinds of assumptions about block execution that cause this to break in the first place.)

Expected/fixed:
tangram-thu may 12 2016 18-46-42 gmt-0400 edt

Observed/broken:
tangram-thu may 12 2016 18-47-11 gmt-0400 edt

### DIFF
--- a/bubble-wrap.yaml
+++ b/bubble-wrap.yaml
@@ -682,32 +682,22 @@ styles:
 
     building-grid:
         base: polygons
-        lighting: vertex
+        lighting: false
         mix: [hsv, scale-buildings]
         texcoords: true
         shaders:
             uniforms:
                 u_tex_grid: building-grid
             defines:
-                WALL_TINT: vec3(1., 3., .800) # modifies roof color HSV
+                WALL_TINT: vec3(1., 3., .800)
             blocks:
-                # applied in vertex shader when using vertex lighting
                 color: |
                     if (dot(vec3(0., 0., 1.), worldNormal()) < 1.0 - TANGRAM_EPSILON) {
                         // If it's a wall
                         color.rgb = hsv2rgb(rgb2hsv(color.rgb) * WALL_TINT);
-                        color.a = 1.0;
-                    } else {
-                        // it's a roof, use vertex color without texture
-                        // using color alpha to differentiate roof/wall
-                        color.a = 0.0;
+                        color.rgb = mix(color.rgb, vec3(0.),
+                                        texture2D(u_tex_grid, v_texcoord).a);
                     }
-                # applied in fragment shader
-                filter: |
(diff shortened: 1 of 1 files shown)


bcamper:
cc @nvkelso @sensescape @blair1618 @hjanetzek

hjanetzek:
I guess it's a bug in the filter shader block that became apparent now: It should be using `color` instead of `v_color` to have `v_lighting` applied. Testing on ES this seems to work correctly.

Just realized that color-block moved to fragment shader. I think it was good before where color-block was used to modify the vertex color and filter-block to modify fragment-color. Or can we have both  `vertex_color` and `fragment_color`?

bcamper:
The `color` block was moved to the fragment shader to solve a different issue, where a procedural texture could be lit with vertex lighting. However, it may turn out this is not as important or common as other cases like this one.

The reason I didn't just update `v_color` is that everything is running in the fragment shader, so keeping the extra bits like encoding surface in `color.a` just adds unnecessary steps.

I'd like to think more about reverting to the old model where `color` is run in the vertex shader, and to hear other people's thoughts.

In the meantime this PR was just to make it compatible with current behavior. We can revert or update again if shader block behavior changes.

nvkelso:
+1

bcamper:
This change should only be deployed in Eraser Map with a version of ES that has the updated lighting block behavior -- @tallytalwar do you know when that will be integrated into Eraser Map / Android SDK?

tallytalwar:
@bcamper the word on the slack is EOD today tangram-0.2.2 should be in EM.

bcamper:
Great, merging this now and then we can update the style at the same time as the library.

PR #186, opened 2016-07-25
Add mz_dash_line data layer

sarahcodes100:
Adds mz_dash_line layer for use in the multi-modal route preview. If we like this, I'll update cinnabar and refill too.

screenshot_20160725-154519

screenshot_20160725-154435

cc/ @souperneon how do you feel about these lines?

### DIFF
--- a/bubble-wrap.yaml
+++ b/bubble-wrap.yaml
@@ -406,6 +406,10 @@ sources:
 #        url: https://gist.githubusercontent.com/anonymous/30c6c1a75c168d91d90c/raw/92bfe55e622766d250b1f2f5d17bdc7c26acb956/map.geojson
 #        # local sf trip
 #        # url: https://gist.githubusercontent.com/anonymous/9a610ebda6fe4be7bccc/raw/8d217e43f2412d48d01534ba115f1e42dac72e68/map.geojson
+#    # Dashed route line
+#    mz_dash_line:
+#        type: GeoJSON
+#        url: https://gist.githubusercontent.com/anonymous/d73b851c64c3e5fbfc2754aa32f44c10/raw/938ae435776e176919c4797bed1465a92e403ef3/map.geojson
 #    # Transit route line
 #    mz_route_line_transit:
 #        type: GeoJSON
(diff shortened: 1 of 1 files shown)


souperneon:
These dashes look a little weird and broken. I'm not sure how much we can edit the appearance but make them more evenly spaced, making them so they have rounded edges and making each dash smaller would be nice? @sarahlensing @nvkelso what do you think?

mjcunningham:
ccing @blair1618

matteblair:
Rounded corners aren't part of the `dash` feature, sorry. You could suggest it on the tangram repo but it will take some work.

The distortion of stripes is something we've seen before - it seems to be worse on thick lines and straight lines, and this hits both of those cases :)

The problem comes down to a detail in the way that line corners are constructed in tangram-es. I'm writing up an issue in the tangram-es repo to address this. In the meantime, these artifacts might be reduced if you try a different `join` type for these lines, like `bevel` or `round`.

sarahcodes100:
I tested `bevel` and `round` and don't notice a reduction in the artifacts. Do we want to forge ahead with what we currently have for a first version and then circle back when rounded corners are added and distortion is fixed? Or hold off on any dash until then?  cc @souperneon @mjcunningham

mjcunningham:
Let's hold off until the distortion is fixed.

nvkelso:
It'll be awhile until any rendering changes in Tangram to be addressed (it's a hard problem). I'm going to merge this now to "start where you are" and unblock the mobile team.

I added an example location to the commented out source layers list, renamed some of the additions to improve consistency, and thickened the dashed line by zoom.

nvkelso:
Released in https://github.com/tangrams/bubble-wrap/releases/tag/v4.1.0.

mjcunningham:
@nvkelso are you saying we should use it in the app? What do your changes look like?

nvkelso:
Yes, we should use in the app and improve it later once Tangram revs.

My changes only thin and thicken the dashed line width per zoom so it matches the transit color lines better.

PR #192, opened 2016-08-16
Fix typos in park and school filters

matteblair:


nvkelso:
Good catch! I think there might be more of these, like for buildings and other landuse tiers?

matteblair:
Could be - I didn't do a thorough search, these just showed up with warnings in tangram-es.

nvkelso:
Looks like these were the only 2 instances.

nvkelso:
Released in https://github.com/tangrams/bubble-wrap/releases/tag/v4.1.1.

PR #194, opened 2016-08-18
Fix precision issues in building shader in some contexts

matteblair:
Re: https://github.com/mapzen/eraser-map/issues/722

In theory this edit should not change the shader output, but in some contexts there appear to be some liberties taken with floating point arithmetic ;)

This resolves the issue above in tangram-js on Edge and may resolve the issue in tangram-es on Sailfish OS (I can't test that one myself).

### DIFF
--- a/bubble-wrap.yaml
+++ b/bubble-wrap.yaml
@@ -852,7 +852,7 @@ styles:
                 WALL_TINT: vec3(1., 3., .800)
             blocks:
                 color: |
-                    if (dot(vec3(0., 0., 1.), worldNormal()) < 1.0 - TANGRAM_EPSILON) {
+                    if (dot(vec3(0., 0., 1.), worldNormal()) < .5) {
                         // If it's a wall
                         color.rgb = hsv2rgb(rgb2hsv(color.rgb) * WALL_TINT);
                         color.rgb = mix(color.rgb, vec3(0.),


bubble-wrap.yaml
```diff
@@ -852,7 +852,7 @@ styles:
                 WALL_TINT: vec3(1., 3., .800)
             blocks:
                 color: |
-                    if (dot(vec3(0., 0., 1.), worldNormal()) < 1.0 - TANGRAM_EPSILON) {
+                    if (dot(vec3(0., 0., 1.), worldNormal()) < .5) {
```
tallytalwar, line 855: Interesting.

`0.5` is too sharp of a change though. This means if the normal of the top of a building is 61 degrees of from the world normal (for any slanting building tops), it will put a grid texture on the building top.

When I tried this on play on Edge, replacing the `TANGRAM_EPSILON` to `0.001` worked for me.

matteblair, line 855: There are no slanting building tops in tangram. This is the simplest way to distinguish between vertical and horizontal vectors.

tallytalwar, line 855: Ohh yeah every point is extruded with the same height. My bad.

nvkelso:
@blair1618 Can you include some before/after screenshots with this, please?

matteblair:
Before:

before-building-grid-fix

After:

after-building-grid-fix

(Screenshots from MS Edge on Windows 10)

nvkelso:
Tangram Frame to the rescue!

Confirmed it still looks fine under macOS via:
- http://tangrams.github.io/tangram-frame/?&url=https://raw.githubusercontent.com/tangrams/bubble-wrap/4676dacde5c762a7ba95828a821cf466bac2e4e1/bubble-wrap.yaml#15/40.7053/-74.0097

PR #195, opened 2016-08-18
Fix some font styling typos

matteblair:
tangram-es produced warnings for a few spots where draw rules were missing a 'font' block.


nvkelso:
LGTM, thanks!

PR #198, opened 2016-09-10
Provide global for switching buildings extrusion

tallytalwar:
We need this for sdk control for tizen.

@bcamper, what you suggested was to put a global for the draw rule parameter, but since `footprint` and `extrude` both have the same named draw rule, if we set `visible` draw rule for `extrude` sublayer to the global, it will not be respected, since `footprint` sublayer's `visible` draw rule is always `true`.

If we wish to apply the global property to the `visible` draw rule of the `extrude` sublayer, we will also have to make sure it has a unique draw rule name and not share it with other sublayers under `buildings` layer.

### DIFF
--- a/bubble-wrap.yaml
+++ b/bubble-wrap.yaml
@@ -361,9 +361,8 @@ global:
     building1: [0.784, 0.784, 0.784]        # building
     building2: [.860, .860, .860]           # building stroke
     building_o:  5                          # building stroke order
-    building_e: true                        # building stroke order
-    #building extrusion toggle
-    building_extrude: true
+    building_e: true                        # building stroke order early
+    building_extrude: false                 # building extrusion toggle
 
 
 textures:
@@ -2694,7 +2693,6 @@ layers:
         # building footprints, pre-extrusion
         footprints:
             filter:
-                $zoom: { max: 18 }
                 any:
                     # show footprints for buildings at least one zoom level before they will be extruded
                     - { $zoom: [13], area: { min: 50000 } }
@@ -2714,6 +2712,14 @@ layers:
                     visible: true
                 lines:
                     visible: true
+
+            footprints-late:
+                filter: { $zoom: { min: 18 } }
+                draw:
+                    polygons:
+                        visible: function() { if(global.building_extrude) { return false; } else { return true; } }
+                    lines:
+                        visible: function() { if(global.building_extrude) { return false; } else { return true; } }
 
         # 3d buildings
         extrude:
--- a/bubble-wrap.yaml
+++ b/bubble-wrap.yaml
@@ -362,7 +362,9 @@ global:
     building2: [.860, .860, .860]           # building stroke
     building_o:  5                          # building stroke order
     building_e: true                        # building stroke order early
-    building_extrude: false                 # building extrusion toggle
+    building_extrude: true                  # building extrusion toggle
+    building_extrude_height: |              # building extrude height logic
+        function() { return feature.height || 20; }
 
 
 textures:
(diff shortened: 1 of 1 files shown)


nvkelso:
@tallytalwar Can you make this a general boolean at the end of the [buildings globals section](https://github.com/tangrams/bubble-wrap/blob/6d5ad2ef8e4f7d7a6167f4c74e281bd0fa2a63d4/bubble-wrap.yaml#L364-L367) instead, please? This doesn't feel specific to the SDK, just a general scene option.

bubble-wrap.yaml
```diff
@@ -2692,7 +2695,6 @@ layers:
         # building footprints, pre-extrusion
         footprints:
             filter:
-                $zoom: { max: 18 }
```
nvkelso, line 2695: This was added way back when because if buildings are extruded we don't want to draw them all twice (which is a performance hit). Removing it here seems to allow buildings (footprints) to show up later when the extruded volumes aren't shown, but introduces a performance hit?

tallytalwar:
Sure can do that.

bubble-wrap.yaml
bcamper: Right I don't believe this should be removed.

matteblair: Strictly speaking, this will negatively impact performance at zooms above 18. However our rendering budget is pretty generous at these zoom levels because there's relatively little geometry on screen.

matteblair: The alternative would be to continue the pattern of area/volume thresholds for footprints and extrusions for higher zoom levels.

bcamper: Right, I would like to not double-draw anything... that should be avoidable by updating these filters appropriately.

matteblair: Noted with new issue: https://github.com/tangrams/bubble-wrap/issues/200

tallytalwar: K, I am going to this change out. @nvkelso can you prioritize https://github.com/tangrams/bubble-wrap/issues/200? We plan to send this over for tizen dev asap.

Thanks

tallytalwar:
@nvkelso Updated with your suggestion.

nvkelso:
@tallytalwar As written this removes all building footprints at zoom 18 if the new global is set to false. Instead we'd want to see building footprints past zoom 17 (the footprints filter is set to max 17 now). I'll have a look to fix right now.

nvkelso:
Now you see them at zoom 17:
screen shot 2016-09-16 at 09 43 38

Now you don't at zoom 18:
screen shot 2016-09-16 at 09 43 43

bcamper:
Yes this should be fixed. Let me know if you need help.

nvkelso:
Here it is fixed via 5c7d9df474ebe74f2c3b696410896005cbbd343b at zoom 18:

screen shot 2016-09-16 at 09 54 07

nvkelso:
@bcamper @tallytalwar This looks good to me now. Can you review the code, please? If it looks good, go ahead and merge. 

Also, do we still need #200? Maybe to refine filters even more to avoid even more double drawing? But I don't think that's needed for the work next week?

bcamper:
I'll take a look, but we definitely do not want to ship anything with
double drawing.

bubble-wrap.yaml
```diff
@@ -2713,8 +2713,17 @@ layers:
                 lines:
                     visible: true
 
+            footprints-late:
```
tallytalwar: This looks good to me.

bcamper: JS functions are not allowed in the `visible` property. 

I am rewriting the surrounding footprint/extrude logic to reduce computation at the same time. Will push in a bit.

tallytalwar: Ahh I always forget not all properties support JS function.

tallytalwar:
Ok this works (in play)

However there might be some bug related to functions in filter in ES, and no extrusion happens. Testing now..

matteblair:
@tallytalwar I've tracked down the bug(s) causing this and I'm correcting them now.

matteblair:
One problem down: https://github.com/tangrams/tangram-es/pull/979

nvkelso:
@blair1618 and @tallytalwar is this ready to merge now? I think the YAML here is good.

(Merging here isn't same as cutting a new release: Eraser Map and Mapzen.js only refer to released versions of the style so we have some leeway here – but I think for your demos you want things merged into gh-pages branch?)

tallytalwar:
Yes, `gh-pages` branch works for us.

We still have to resolve an issue on `tangram-es`, I guess as soon as that is resolved, we can merge this.

nvkelso:
_passes @tallytalwar the merge button_

matteblair:
Last problem resolved: https://github.com/tangrams/tangram-es/pull/980

After that is merged this change to the scene file will work beautifully.

PR #199, opened 2016-09-12
Use a 'global' to set building extrusion visibility

matteblair:
Resolves #198 

Originally I wanted to use a global for the `extrude` property. That works fine when a client application only wants to turn extrusion to `false`, but then to restore the original behavior the client code would need to know the function that was previously used for extrusion. Since we draw building footprints in a separate layer anyway, it's easier just to set the visibility of the layer called "extrude".

### DIFF
--- a/bubble-wrap.yaml
+++ b/bubble-wrap.yaml
@@ -363,6 +363,9 @@ global:
     building_o:  5                          # building stroke order
     building_e: true                        # building stroke order
 
+    # building extrusion
+    building_extrude_on: true
+
 
 textures:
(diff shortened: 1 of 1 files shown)


bubble-wrap.yaml
```diff
@@ -2692,7 +2695,6 @@ layers:
         # building footprints, pre-extrusion
         footprints:
             filter:
-                $zoom: { max: 18 }
```
tallytalwar, line 2695: Incorporating this in my PR.

PR #201, opened 2016-09-13
Couple of globals were missing resulting in warning in tangram

tallytalwar:
@nvkelso Have a look at this, I am not sure if you want these globals to be marked `true` or `false`. I have gone with `true` but that means these will start showing up now, but were not before.


nvkelso:
LGTM, thanks for the fix!

PR #202, opened 2016-09-16
add optional Transit overlay 

nvkelso:
I think this is good enough to merge for now.


nvkelso:
To enable this, set the `global.sdk_transit_overlay` to `true`.

Preview images:

screen shot 2016-09-16 at 10 48 02
screen shot 2016-09-16 at 10 46 47
screen shot 2016-09-16 at 10 46 27

nvkelso:
@bcamper @tallytalwar I'm done with Transit Overlay option. Please review and merge at your leisure (I'll be MIA for most of the rest of the day).

layers/transit-overlay.yaml
```diff
@@ -0,0 +1,399 @@
...
+                                weight: 600
+                                size: 12px
+        subway-early:
+            filter: { is_subway: true, is_train: false, $zoom: { max: 12 } }
+            draw:
+                icons:
+                    visible: false
+                    text:
+                        visible: false
+        halt-early:
```
tallytalwar, line 245: A bunch of these layers including `halt-early`, `platform-early`, `tram-stop-early`, are overriding the usage of `visible: global.sdk_transit_overlay` to always be `true`. Is that intended?

tallytalwar: @bcamper you want to confirm this?

Other than this, this PR looks good to be merged.

bcamper: I'm sure the `visible` override is a mistake.

bcamper:
The screenshots look great :)

But, this is actually a big change to the scene organization because it adds imports to a house style for the first time. I like the start of a transition to a more modular structure, but I don't want to make a change of this magnitude so quickly, especially right before our demo trip next week. Tangram ES supports imports, but we haven't used them in any of our core use cases yet, so we should make sure to test/benchmark before putting a bigger production dependency on the feature.

Let's hold on merging this, and revisit after Mapzencon.

Couple other comments:
- We can simplify the scope of this PR by only having one `import`, for the transit layer. It makes sense to add the transit layer as a separate scene file, because it is truly additive, and can be similarly imported for other styles. This lets us ease Bubble Wrap into using imports more gently, and we can split additional files out in separate PRs.
- For the new files, let's not lump them all in under `layers/`. This is potentially confusing, because not all of those files correspond to scene `layers`. The transit layer itself does, so that makes sense. But `globals.yaml` and `textures.yaml` don't so I'd rather find a different naming scheme (separate folders, different folder name?). I think those are the right places to start modularizing though (global/SDK-level flags, and "icon packs" we can reuse elsewhere).

nvkelso:
Closing this in favor of #203.

layers/transit-overlay.yaml
nvkelso: I'll have a look at the visible overrides in the other PR.

nvkelso:
@bcamper Sure thing about simplified structure for now – other PR looks good.

By the way, Tron uses the same generic `layers/` imports so we should revisit that before publishing. I've added https://github.com/tangrams/tron/issues/110 to track that.

nvkelso:
@bcamper Some background: The reason I went with the imports for more than just the transit layer is @sensescape and I haven't seen consistent availability of globals that are imported into the base YAML file also being available in other imported files. So to make the new `visible: global.sdk_transit_overlay` available there, it needed to go into a file imported that could also be imported into the imported transit file. Is that a Tangram bug?

bcamper:
Hmm, haven't seen behavior like that, the scene merging of imports happens
before any globals are applied, so I don't know what would cause that,
maybe a side effect of another issue? Can you send an example I can test?

In any case, we _don't_ want to import files multiple times into others,
it's just adding redundant work and possibly masking other issues. So let's
definitely not ship anything with that (this or Tron).

PR #204, opened 2016-09-19
Use map background color instead of earth layer

bcamper:
For this style (and some other house styles), we don't need to draw an `earth` layer at all, we can just use the map background color. This reduces a lot of pixel fill volume, which can be a bottleneck for us.

I found an improvement of up to 5fps depending on scene complexity (and this was on non-retina, we should also test displays with higher pixel volume).

Differ tests confirm there's not difference at all in rendering output:
http://tangrams.github.io/differ/?1=https://tangrams.github.io/bubble-wrap/bubble-wrap.yaml&2=https://raw.githubusercontent.com/tangrams/bubble-wrap/background-color/bubble-wrap.yaml&lib1=0.10&lib2=0.10&ignore1&ignore2&go

### DIFF
--- a/bubble-wrap.yaml
+++ b/bubble-wrap.yaml
@@ -288,7 +288,6 @@ global:
     fence: [0.827,0.808,0.780]
     #
     #landuse
-    scene1: '#eee'                          # map background, water
     water1: [0.83, 0.83, 0.83]              # water
     water2: [.75,.75,.75]                   # playa
     water1_o: '#9dc3de'                     # water stroke
@@ -959,7 +958,7 @@ styles:
 
 scene:
     background:
-        color: global.scene1
+        color: [0.870,0.870,0.870]
 
 layers:
     # Map overlays for styling the server response (using special source layer names) for route line, current location, and search result pins
@@ -1108,13 +1107,6 @@ layers:
                 width: 0px
 
     # Basemap styling
-    earth:
-        data: { source: mapzen, layer: earth }
-        draw:
-            polygons:
-                order: global.feature_order
-                color: [0.870,0.870,0.870]
-
     water:
         data: { source: mapzen, layer: water }
         draw:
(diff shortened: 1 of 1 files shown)


bubble-wrap.yaml
```diff
@@ -959,7 +958,7 @@ styles:
 
 scene:
     background:
-        color: global.scene1
+        color: [0.870,0.870,0.870]
```
bcamper: Could be moved into a global, but wasn't before anyway (see old earth layer below).

nvkelso:
This makes me a little nervous as there are sometimes islands that need to stack above water (and the whole Venice under water fiasco). I'll have a look.

bubble-wrap.yaml
nvkelso: Yes, should definitely be a global.

bcamper:
Hm, OK, we should find a way to address with island-specific rules if needed, because what we've been doing is hugely wasteful...

bubble-wrap.yaml
nvkelso: Looks like we should update the `earth1` global to `[0.870,0.870,0.870]` and then reference that global in the background section above.

bcamper:
FWIW I did a quick check on Venice and this view is pixel-identical... but I don't know all the nuanced places that this can crop up.

tangram-1474306092676

nvkelso:
Verified we only set `sort_rank` on default values, or special on ridge and arete lines for labels. The filters for **earth** don't include anything else besides the polygons that are used to generate label placements (and then get dropped).

After the global variable change, see https://github.com/tangrams/bubble-wrap/pull/204#r79443635, we should merge this.

bubble-wrap.yaml
nvkelso: Fixed in https://github.com/tangrams/bubble-wrap/pull/204/commits/8420e66c6820d9030495c9202a96340e2f06ce54.

PR #212, opened 2016-11-21
Use partially evaluated ux_language_text_source functions

hjanetzek:
This change avoids some string allocations when ux_language/fallback is defined, making it slightly more efficient (at least with duktape)


hjanetzek:
Looking at it again now: This change would break the fast-path for language updates I suggested :)

Edit: no this would still work - since on ES our JS StyleContext would need to be recreated after global changes.  

'The first self-transforming map style'!

nvkelso:
I think I understand what's going on in this PR, but requesting that @bcamper gives it some pro-level JavaScript review.

PR #218, opened 2016-12-01
icon refactor

sensescape:
icons renamed to tilezen kind, sdk shields added


nvkelso:
LGTM

matteblair:
Is it possible to pack the sprite sheet such that the image is <= 1024 pixels on both sides? There are Android devices in the wild that don't support textures with dimensions larger than 1024 (the number of such devices is steadily decreasing, but last I checked it was non-trivial).

nvkelso:
@blair1618 We already ensure the texture is <= 1024 in _width_, but don't enforce a similar depth limit at the moment. The unfortunate thing is we actually have enough icons now that we're hitting up against the _1024_ depth because of **valid content**. It's not possible to just clip it. (The white areas in the middle of the spritesheet are actually white shield primitives with transparency.) Temporarily I may be able to manually move things around in Illustrator (order of operations on export) to get below 1024 depth on this spritesheet, but some of the other styles have more content so that approach isn't going to work. Suggestions?

matteblair:
The image is currently just barely over 1024 pixels (1096), it shouldn't take much adjustment. There's still a fair amount of free space in the image, I don't know for sure how your export script works but I bet that if you just re-order some of the sprites you can pack things tighter vertically. A good, simple heuristic is to sort by size before packing.

nvkelso:
@blair1618 I've manually resorted the Illustrator file and all the spritesheets are _just barely_ under `1024` width by `1024` height now. They don't really have room for any other road shields, and maybe another row of small standard map icons. I'll commit those separately to each house style / open PR that Geraldine has.

Filed https://github.com/tangrams/tangram-es/issues/1145 for upstream Tangram ES changes.

nvkelso:
@blair1618 Icon depth is now fixed for Bubble Wrap via https://github.com/tangrams/bubble-wrap/commit/657f64f9706cba1aca2bb518cda391a1465b1695.

PR #221, opened 2016-12-07
depth less than 1024, for matt

nvkelso:
The earlier icon refactor had a spritesheet larger than 1024 px square. This PR fixes that.


matteblair:
You're too kind :)

PR #222, opened 2016-12-07
places refactor

nvkelso:
Running tally of changes in this PR:

- Places layer changes:
  - Refactor `locality` section for legibility and to reduce length (overall complexity), adds better collisions per more population classes
    - Seems to be bug in Tangram around 0px sized icons where text attachment is not always displayed still. Worked around this by moving from icon to text symbolizer at zoom 11, but feels like a hack and duplicates code. Search for `filter: { kind: locality }` in diff for the simple declaration and then in each of the population classes that follow (like `population-10m-up`).
    - This hack should use `text-blend-order` instead of `text`?
  - Refactor `country` and `region` sublayers for same – many more country labels at low zooms (and they have localized names!)
  - Improve hierarchy contrast between locality and other labels (new font treatments, townspot icon sizes, font sizes)
  - neighbourhood refactor (from Tron)
- Add boundary line labels (has some problems in Tangram where the order is sometimes correct, sometimes not if Tangram "flips" the label, or label has space in it then odd line wrap happens (see NY - NJ boundary line)
- Fixed airport runways (the lines were hidden at all zooms, oops).
- Adjusted boundary line widths and colors, and dashing on disputed.
- Fix building color bug around `landuse_kind`
- POIs get new font to improve contrast with other labels


sensescape:
- [x] Fix weird label density & text sizes outside of Paris, seems related to `kind_detail:locality`?
  - http://tangrams.github.io/bubble-wrap/#13/49.0821/1.9873
  - http://localhost:8005/#13/49.0821/1.9873
- [x] Fix too small text labels in same area above for `Us` and other bigger towns, which are `kind_detail:village` with tiny populations.
- [x] Fix similar problem to `kind_detail:locality` but for hamlets in Hong Kong:
  - http://localhost:8005/#13/22.3814/114.2705

screen shot 2016-12-08 at 4 29 56 pm

bubble-wrap.yaml
```diff
@@ -3806,1831 +3893,681 @@ layers:
                         family: Varela
                         # weight: 500
                         fill: global.text_fill
-            text-blend-order:
+            # NOTE: This is a hack for localities because of an icon & text bug, see note below
+            text:
```
sensescape: `text-blend-order`

bubble-wrap.yaml
```diff
@@ -3806,1831 +3893,681 @@ layers:
...
                 text_source: global.ux_language_text_source
                 visible: false    # labels are enabled by each layer below
                 font:
                     family: Varela
-                    # weight: 500
                     fill: global.text_fill
+            # END NOTE
 
-        continent:
```
sensescape, line 3801: This was double specified in places and earth layers, but the data only exists in earth layer as of v1.0 vector tiles.

bubble-wrap.yaml
```diff
@@ -3806,1831 +3893,681 @@ layers:
...
                     #collide: false
                     text:
+                        visible: global.text_visible_populated_places
                         interactive: true
                         buffer: 3px
                         # debug testing
                         #collide: false
+                        font:
+                            size: [[5,9px],[8,10px],[12,11px]]
+                text:
```
sensescape: `text-blend-order`

bubble-wrap.yaml
```diff
@@ -3806,1831 +3893,681 @@ layers:
...
+            #                 anchor: center
+            #
+            # NOTE: So instead we play a shell game with symbolizers per zoom
+            #
+            _icons_later: 
+                filter: { $zoom: { min: 11 } }
+                draw:
+                    icons: 
+                        visible: false
+                    text:
```
sensescape: `text-blend-order`

bubble-wrap.yaml
```diff
@@ -3806,1831 +3893,681 @@ layers:
...
+                        visible: global.text_visible_populated_places
+
+            _text_early: 
+                filter: { $zoom: { max: 11 } }
+                draw:
+                    text:
```
sensescape: `text-blend-order`

bubble-wrap.yaml
```diff
@@ -3806,1831 +3893,681 @@ layers:
...
+                        #color: yellow
+                        #visible: false
+
+            # HACK
+            # NOTE: These should use icon: buffer instead of going visible: false, but there are problems using icon here :(
+            # This fixes Paris area, but causes problems for small places in the USA eg Table Bluff near Eureka, California.
+            funky-village:
+                filter: { kind_detail: [village], $zoom: [11], population: { max: 1000 } }
+                draw:
+                    text:
```
sensescape: `text-blend-order`

bubble-wrap.yaml
```diff
@@ -3806,1831 +3893,681 @@ layers:
...
+                        visible: false
+            funky-isolated_dwelling:
+                filter: { kind_detail: [isolated_dwelling], $zoom: [13] }
+                draw:
+                    text:
```
sensescape: `text-blend-order`

bubble-wrap.yaml
```diff
@@ -3806,1831 +3893,681 @@ layers:
...
+                        visible: false
+            hamlet:
+                filter: { kind_detail: hamlet, $zoom: [13] }
+                draw:
+                    text:
```
sensescape: `text-blend-order`

bubble-wrap.yaml
```diff
@@ -3806,1831 +3893,681 @@ layers:
...
+                        visible: false
+            # END HACK
+
+            # someone in London and Salt Lake City thought neighbourhoods should be tagged place: locality
+            # They were wrong
+            funky-fake-neighbourhoods:
+                filter: { kind_detail: [locality], $zoom: { min: 13 } }
+                draw:
+                    text:
```
sensescape: `text-blend-order`

bubble-wrap.yaml
```diff
@@ -3806,1831 +3893,681 @@ layers:
...
-                            text:
-                                visible: global.text_visible_populated_places
-                                font:
-                                    size: 11px
-                                    fill: global.text_fill
-                                    # stroke: { color: global.text_stroke, width: 4 }
-                z3places-2:
-                    filter: { min_zoom: [3] }
+                                size: [[2,11px],[6,15px],[8,16px],[10,18px],[12,18px],[13,0px]]
+                    text:
```
sensescape: `text-blend-order` and etc for all the population classes below

bubble-wrap.yaml
nvkelso: Fixed in 412c008f32a8ada1c0e12e4c2cea9773cbec2e6a.

nvkelso: And in 6a39aac9d8aef70a1f324a6fdc04f77d82e53d40.

nvkelso:
LGTM, thank you!

PR #223, opened 2016-12-07
WIP: interactive icons

nvkelso:

PR #224, opened 2016-12-07
deliberate interactive (and less for prod)

nvkelso:
Takes a different approach than #223 by adding new _internal-only_ globals for interactivity levels and then the individual feature / layer interactivity based on those globals.

- keep POIs interactive
- keep some UX icons like search results and default pins interactive
- disable interactivity on everything else (roads, buildings, transit lines, etc)

### DIFF
--- a/bubble-wrap.yaml
+++ b/bubble-wrap.yaml
@@ -167,8 +167,8 @@ global:
     sdk_shield_color: [0.506,0.192,0.169]
     sdk_shield_text_color: white
     #
-    interactive_dev: false
-    interactive_prod: true
+    # enable interactivity for key features
+    interactive: true
     #
     # default order for basemap features
     feature_order: function() { return feature.sort_rank; }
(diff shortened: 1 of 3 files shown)


ecgreb:
@nvkelso we will give this a try today in the Android SDK and Eraser Map

nvkelso:
@bcamper has even better ideas for enabling newer Tangram's introspection features via a config for debug – so more inline with @tallytalwar PR, but keeping some of my `interactive: true` statements for "prod".

bubble-wrap.yaml
```diff
@@ -167,8 +167,8 @@ global:
     sdk_shield_color: [0.506,0.192,0.169]
     sdk_shield_text_color: white
     #
-    interactive_dev: false
-    interactive_prod: true
+    # enable interactivity for key features
+    interactive: true
```
bcamper, line 171: This could also be `ux_interactive` or `sdk_interactive` (what is naming criteria for ux vs. sdk?) for namespacing and consistency.

tallytalwar: `sdk_interactive` is more inline with other `sdk_*` globals we have.

nvkelso: originally my thinking was `ux_*` was for user interaction / app level and `sdk_*` for changing parts of the base map. But it seems like `sdk_*` is generally better (and one less name space). So I'd name this `sdk_interactive`.

nvkelso:
Looks like this still needs to pair with html/JS changes to enable the newer Tangram introspection before merging.

bcamper:
Yep those are there now.

bcamper:
Should be ready to go!

nvkelso:
The new behavior is:

1. default
  - only a minority of "POI" like features are interactive in JS or ES:
    - all POIs
    - some UX icons like search results and default pins
  - but because interactive_debug enabled in the HTML rig in this repo:
    - all features are interactive (except the map background, e.g. "earth"), via Tangram introspection (not scene file interactive flags)
    - this preserves the debug functionality needed to develop the style on the desktop!
2. when interactive_debug is disabled for JS via the HTML rig using the GUI:
  - then it matches the default ES behavior

PR #225, opened 2016-12-13
pois layer refactor

nvkelso:
Uses the following pattern:

- All pois visible by default
- Some pois source sprite from something other than kind
- Many pois should be hidden early
- Some pois should receive special icon or text styling
- Most custom overrides have been removed
- Most nested layers have been removed


nvkelso:
Final commit restored highway exits!

PR #230, opened 2016-12-27
Add transit line shields

### LINKED ISSUE #209: Disable road shields when transit overlay enabled
New road shields are landing, and on by default. But those might not be helpful to transit (and transit could show shields for the lines instead).

There's a global to disable them. Let's graduate that to a official "API" global (change the name, document in `cartography-docs`), and update the logic in the layer switcher.

/cc @irees.

### LINKED ISSUE #210: Add transit line shields in transit overlay
As we remove road shields in #209 when in transit overlay mode, we should also show shields for transit lines.

nvkelso:
fixes #210 and 
fixes #209 for no road shields in transit overlay

### DIFF
--- a/bubble-wrap.yaml
+++ b/bubble-wrap.yaml
@@ -192,6 +192,7 @@ global:
     # SHIELDS
     sdk_shield_color: [0.506,0.192,0.169]
     sdk_shield_text_color: white
+    sdk_shield_visible: true
     #
     # enable interactivity for key features
     interactive: true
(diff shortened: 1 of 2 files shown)


nvkelso:
This is a quick hack re-using the SDK shield assets. Later we should use a custom transit shield (https://github.com/tangrams/bubble-wrap/issues/231).

15/40.7091/-74.0029

screen shot 2016-12-27 at 15 38 22

PR #240, opened 2017-03-14
add left right boundary labels

nvkelso:
Requires Tangram v0.12, which isn't released yet.


bcamper:
It's released now ;)

nvkelso:
And Tangram ES v0.6 now supports this, too :)

PR #252, opened 2017-07-10
Copied over generic shield sprite coordinates from bubble-wrap-style.yaml

tallytalwar:
cc. @nvkelso


nvkelso:
Looks like the no-labels version also has this funk (my bad), and the spritesheet definition has other inconsistencies. I'll review with @sensescape IRL (there is also a -2 in one of the positions so needs a wider review).

nvkelso:
Thanks for finding this!

Closed in favor of more extensive changes in 2ff01e26847cc42e75b6b498757d827b936a41ea.

PR #262, opened 2018-01-02
updated icons and more shields

sensescape:


bubble-wrap-style.yaml
```diff
@@ -2806,31 +2817,430 @@ layers:
...
+                                      /^D .*$/.test(feature.ref))
+                            }
                         draw:
                             mapzen_icon_library:
-                                size: [[6,[25px,14px]],[13,[28px,15px]],[15,[34px,18px]]]
-                    width_5char:
-                        filter: function() { return (feature.shield_text.length === 5) }
+                                sprite: |
+                                    function() {
+                                        var text_len = feature.shield_text.length ? feature.ref.length;
```
bcamper: This is an incorrect ternary syntax:

`feature.shield_text.length ? feature.ref.length;`

you need a `:` with the alternate value, or I think what you actually want to do is maybe this?

`feature.shield_text.length || feature.ref.length`

bcamper: This is causing the current v9 Bubble Wrap (released from this un-merged branch? I can't tell) to fail in various region/zoom combinations.

bubble-wrap-style.yaml
```diff
@@ -2806,31 +2817,430 @@ layers:
...
+                                        return ( 'FR:D-' + feature.text_len + 'char');
```
bcamper: Related to above, not sure this is what you actually mean here? Is there a `text_len` property on the feature itself? I assume you want to use the variable you just defined above, like:

`return ( 'FR:D-' + text_len + 'char');`

bubble-wrap-style.yaml
```diff
@@ -2806,31 +2817,430 @@ layers:
...
+#                    width_5char:
+#                        filter: function() { return (feature.shield_text.length === 5) }
+#                        draw:
+#                            mapzen_icon_library:
+#                                size: [[6,[22px,17px]],[13,[25px,19px]],[15,[30px,22px]]]
+
+                international-shields:
+                    filter: |
+                        function() {
+                            return feature.shield_text &&
```
bcamper: You need to make the precedence clearer here, cases where `feature.shield_text` is null are getting through. Should be:

```
return feature.shield_text && (
  /^AM:AM.*$/.test(feature.network) ||
   ...
  /^ZA:R.*$/.test(feature.network)
 )
```

bcamper: fixed in https://github.com/tangrams/bubble-wrap/pull/262/commits/33743924702df949c53fee76bf59e38c4d1ad099

bubble-wrap-style.yaml
bcamper: fixed in https://github.com/tangrams/bubble-wrap/pull/262/commits/33743924702df949c53fee76bf59e38c4d1ad099

bubble-wrap-style.yaml
bcamper: fixed in https://github.com/tangrams/bubble-wrap/pull/262/commits/33743924702df949c53fee76bf59e38c4d1ad099

bcamper:
With my most recent commits, I'm not finding any errors right now when trying a broad range of locations (that doesn't mean they aren't there though!).

Some of the same fixes need to be applied to Refill (and probably others?).

bubble-wrap-style.yaml
```diff
@@ -3174,12 +3175,8 @@ layers:
                     france-a-and-n-road-ref-shields:
                         filter: |
                             function() {
-                                return ((feature.ref && ! feature.shield_text) &&
-                                        /^A .*$/.test(feature.ref) ||
-                                        /^N .*$/.test(feature.ref) ) ||
-                                    ((feature.ref && feature.network == 'e-road') &&
-                                    /^A .*$/.test(feature.ref) ||
-                                    /^N .*$/.test(feature.ref) )
+                                return ((feature.ref && !feature.shield_text) && (/^A .*$/.test(feature.ref) || /^N .*$/.test(feature.ref))) ||
```
bcamper: I'm honestly not entirely sure what we're trying to test for here, so the syntax now works (I think the old one was short-circuiting some logic), but it may not be correct...

nvkelso: This is for A-roads and N-roads in Europe (like in France).

nvkelso:
Merging this now to reflect what's been released on [Nextzen as v9](https://www.nextzen.org/carto/bubble-wrap-style/9/bubble-wrap-style.zip) which is based on https://github.com/tangrams/bubble-wrap/pull/262/commits/beb020b69b4dfb0a6f310e6c92a3ac1996630eac. Since then @bcamper has made a couple bug fixes, so after merging we should release a `v9.0.1` bug fix that may also address visual changes as requested in https://github.com/tangrams/bubble-wrap/issues/264.

PR #270, opened 2018-03-28
shields polish: USA and international

nvkelso:
* **Shield sizes** - resize the spritesheets to be at correct sizes, use texture size and then % off that via new Tangram functionality
* **Shield text positioning** - for text that's offset from sprite center (like Florida)
* **Icon sizes** - use texture size and then % off that via new Tangram functionality
* **Tilezen vector tiles v1.5** schema changes (new pois, landuse polygons, building entrances)


bubble-wrap-style.yaml
```diff
@@ -2374,14 +2245,14 @@ layers:
...
                             fill: [0.308,0.308,0.308]
-                            size: [[7, 7px], [12, 8px], [15, 11px]]
+                            size: [[7, 7px], [13, 9px], [15, 10px]]
                             stroke: { color: global.text_stroke, width: 0px }
 
             default_priority_highway:
                 filter:
                     all:
                         - kind_detail: motorway
-                        - $zoom: { min: 9 }
```
nvkelso, line 2384: Please show before/after images, this seems like a big change.

bubble-wrap-style.yaml
```diff
@@ -2176,7 +2047,7 @@ layers:
                 draw:
                     text-blend-order:
                         priority: 59
-                        visible: global.text_visible_minor_road_e
```
nvkelso, line 2179: Please show before/after images, I remember this being for ~ Tangram JS v0.8 limitations that were later fixed so relying on the v0.15 engine to draw better by default probably makes sense here.

bubble-wrap-style.yaml
```diff
@@ -2429,42 +2300,33 @@ layers:
...
-#                        if( feature.shield_text ) {
-#                            return (feature.shield_text.length === 3)
-#                        } else {
-#                            return (feature.ref.length === 3)
-#                        }
-#                    }
-#                draw:
-#                   mapzen_icon_library:
-#                       size: [[6,[25px,14px]],[13,[27px,15px]],[15,[33px,18px]]]
+            width_1char:
```
nvkelso: Not sure we need this section / not sure what it does (it was commented out before -- and the new version only filters, doesn't draw).

bubble-wrap-style.yaml
```diff
@@ -2544,35 +2406,7 @@ layers:
...
-                        filter: function() { return (feature.shield_text.length === 4) }
-                        draw:
-                            mapzen_icon_library:
-                                size: [[6,[25px,18px]],[13,[27px,20px]],[15,[35px,25px]]]
-                    width_5char:
-                        filter: function() { return (feature.shield_text.length === 5) }
-                        draw:
-                            mapzen_icon_library:
-                                size: [[6,[26px,18px]],[13,[30px,20px]],[15,[38px,25px]]]
+                                offset: [0px, 0px]
```
nvkelso: I think this is a default so we can probably just remove?

bubble-wrap-style.yaml
```diff
@@ -2631,6 +2440,34 @@ layers:
                 #             # you need to match any custom shield to the vector tile `network` values
                 #             #priority: 48
                 #             visible: global.sdk_road_shields
+                US-AL:
+                    # Match Alabama state highways: `US:AL`
+                    filter: |
```
nvkelso: Would be better to state general USA matching list (like we do with international) so we don't need to restate the draw texture sprite bit repetitively... and then only the states with custom text offsets should get a sublayer.

bubble-wrap-style.yaml
```diff
@@ -2644,134 +2481,376 @@ layers:
                             #priority: 48
                             text:
                                 offset: [0px, 1px]
-                                font:
-                                    fill: [0.308,0.308,0.308]
-                                    size: [[7,7px],[13,9px],[15,10px]]
-                    motorway:
```
nvkelso, line 2650: I remember California being special (like France) where some of the state roads needed different shield visibility than default. Please show before/after screenshots.

bubble-wrap-style.yaml
```diff
@@ -2644,134 +2481,376 @@ layers:
...
+                    draw:
+                        mapzen_icon_library:
+                            texture: mapzen_icon_library_shields_usa
+                            sprite: function() { return (feature.network + '-' + feature.shield_text.length + 'char'); }
+                            #priority: 48
+                            text:
+                                offset: [0px, 1.5px]
+
+                US-SD:
+                    # Match Pennsylvania state highways
```
nvkelso: Comment state Penn. doesn't match layer name (South Dakota).

bubble-wrap-style.yaml
```diff
@@ -3173,11 +3226,55 @@ layers:
...
+                            }
+                        draw:
+                            mapzen_icon_library:
+                                sprite: |
+                                    function() {
+                                        return ( 'CA:BC-' + feature.shield_text.length + 'char');
+                                    }
+                                text:
+                                    offset: [0px, 2px]
+                            width_1char:
```
nvkelso: Probably don't need this sizing anymore (rely on texture size).

bubble-wrap-style.yaml
```diff
@@ -3218,6 +3316,63 @@ layers:
...
+                            mapzen_icon_library:
+                                sprite: |
+                                    function() {
+                                        return ( 'BAB-' + feature.ref.length + 'char');
+                                    }
+                                text:
+                                    font:
+                                        fill: [1.00,1.00,1.00]
+                            #this is not working, but color is working, see berlin
+                            width_1char:
```
nvkelso: Probably don't need this sizing anymore (rely on texture size).

bubble-wrap-style.yaml
```diff
@@ -3238,6 +3393,140 @@ layers:
                                     function() {
                                         return feature.ref.replace("MEX ", "");
                                     }
+                        #sizing is not working here either
+                        width_1char:
```
nvkelso: Probably don't need this sizing anymore (rely on texture size).

themes/bubble-wrap-icons.yaml
```diff
@@ -723,299 +723,300 @@ textures:
     mapzen_icon_library_shields_usa:
         url: images/bubble-wrap-shields-usa@2x.png
         filtering: mipmap
-        density: 5
+        density: 2
```
nvkelso: The default icons density should also be 2? 

- https://github.com/tangrams/bubble-wrap/pull/270/files#diff-131e16e219854a71b9922916954b6624L45

themes/label-5.yaml
```diff
@@ -26,11 +26,11 @@ global:
     text_visible_trunk_primary_route:   true        # true by default
     text_visible_trunk_primary_e:       false     # false for default, false for more
     text_visible_secondary:             true
-    text_visible_secondary_e:           false     # false for default, false for more
+    text_visible_secondary_e:           true     # false for default, false for more
     text_visible_tertiary:              true
     text_visible_tertiary_e:            false     # false for default, false for more
     text_visible_minor_road:            true
-    text_visible_minor_road_e:          true
+    text_visible_minor_road_e:          false
```
nvkelso, line 33: If these are no longer referenced (other code changes removed them, but require further discussion) consider removing them here from list of globals.

tallytalwar:
Request to put `%` usage in a separate commit or ideally in a separate PR, to make it easier for development comparison for ES.

bubble-wrap-style.yaml
```diff
@@ -433,8 +433,10 @@ fonts:
 
 sources:
     mapzen:
-        type:  MVT #TopoJSON
-        url:  https://tile.nextzen.org/tilezen/vector/v1/512/all/{z}/{x}/{y}.mvt
+        type: MVT #TopoJSON
+        #url:  https://tile.nextzen.org/tilezen/vector/v1/512/all/{z}/{x}/{y}.mvt
```
nvkelso: Reset to prod before merging.

nvkelso:
Consider splitting USA and international shields into separate icon bundles that are like theme imports.

nvkelso:
@sensescape and @bcamper this is ready for review, finally! :)

bubble-wrap-style.yaml
```diff
@@ -2342,7 +2224,7 @@ layers:
                             }
                         }
                     sprite_default: generic_shield_5char
-                    size: [[7,[90%,90%]],[12,[100%,100%]],[15,[110%,110%]]]
+                    size: [[7,[75%,75%]],[12,[100%,100%]],[15,[110%,110%]]]
```
bcamper, line 2227: Minor note, with %-based sprite scaling you should be able to write this with a single value since it's the same in both dimensions, e.g. 

`size: [[7, 75%],[12, 100%],[15, 110%]]`

bubble-wrap-style.yaml
```diff
@@ -2374,60 +2256,102 @@ layers:
...
-                        - $zoom: { min: 9 }
                 draw:
                     mapzen_icon_library:
                         priority: 47
                         #color: yellow
                         visible: global.sdk_road_shields
-
+                early:
+                    filter:
+                        - $zoom: { max: 7 }
```
bcamper, line 2274: Another minor thing, but in all these cases where we have a single filter condition, it's more succinct to not put it in an array, e.g.

```
filter:
  - $zoom: { max: 7 }
```

can just be

```
filter:
  $zoom: { max: 7 }
```

In the former, the engine is creating an unnecessary array filter for multiple conditions, even though there's only one (unlikely we'd ever see any real-world performance impact from this though).

bubble-wrap-style.yaml
```diff
@@ -1910,12 +1791,13 @@ layers:
                             visible: global.text_visible_secondary
                             text_source: global.ux_language_text_source_road_ref_and_name
                             font:
-                                # need a fill color here?
                                 fill: [0.15,0.15,0.15]
                                 size: 16px
                                 stroke: { color: [1.00,1.00,1.00], width: 3px }
             tertiary:
-                filter: { kind_detail: [tertiary, tertiary_link] }
+                filter:
```
bcamper, line 1798: The original syntax here was more succinct, e.g. could just be:

`filter: { kind_detail: [tertiary, tertiary_link, unclassified] }`

`all` or `any` don't have any effect when there's only a single condition. It's slightly more efficient for the engine to parse without it.

bubble-wrap-style.yaml
```diff
@@ -2374,60 +2256,102 @@ layers:
...
                             weight: 500
                             fill: [0.308,0.308,0.308]
-                            size: [[7, 7px], [12, 8px], [15, 11px]]
+                            size: [[7, 7px], [13, 9px], [15, 10px]]
                             stroke: { color: global.text_stroke, width: 0px }
 
             default_priority_highway:
                 filter:
                     all:
+                        - kind: highway
```
bcamper, line 2265: See note above on filter simplification, this could just be:

`filter: { kind: highway, kind_detail: motorway }`

`all` is implied when there is more than one filter condition.

bubble-wrap-style.yaml
```diff
@@ -2374,60 +2256,102 @@ layers:
...
+                        - $zoom: { max: 7 }
+                    draw:
+                        mapzen_icon_library:
+                            visible: false
             default_priority_trunk:
                 filter:
                     all:
-                        - kind_detail: [trunk]
-                        - $zoom: { min: 10 }
+                        - kind: major_road
```
bcamper, line 2281: See filter simplification notes above.

bubble-wrap-style.yaml
```diff
@@ -3876,6 +3085,29 @@ layers:
                         style: italic
                         size: 10px
                         stroke: { color: global.text_stroke_address, width: 4 }
+        entrance-labels:
+            filter:
```
bcamper, line 3129: This *might* be clearer as:

```
filter:
  all:
    - $zoom: { min: 20 }
    - kind: entrance
    - function() { return global.text_visible_address; }
```

bcamper: While `all` is usually implied, the `function` requires some alternate syntax here.

bubble-wrap-style.yaml
```diff
@@ -3876,6 +3085,29 @@ layers:
...
+                $zoom: { min: 20 }
+                any:
+                    - kind: entrance
+                all:
+                    - function() { return global.text_visible_address; }
+            draw:
+                mapzen_icon_library:
+                    size: [[13, [95%,95%]], [16, [100%,100%]], [18, [105%,105%]]]
+                    sprite: generic #function() { return feature.kind; }
+                    sprite_default: generic
```
bcamper: If sprite is always `generic` (line above), then `sprite_default` is extraneous here.

bubble-wrap-style.yaml
```diff
@@ -3366,7 +2563,19 @@ layers:
...
-                        width: [[15, 0px], [16, 0.5px], [17, 1.0px], [18, 2.0px]]
+                        width: [[14, 0px], [15, 0.2px], [16, 0.5px], [17, 1.0px], [18, 2.0px]]
+            major_track:
+                filter:
+                    any:
+                        - { kind_detail: track, surface: [paved, asphalt, concrete, gravel] }
+                        - { kind_detail: track, tracktype: [grade1, grade2] }
+                draw:
+                    lines:
+                        color: [1.0,1.0,1.0]
```
bcamper, line 2614: IMO it's slightly clearer to just say `color: white` for these cases, but totally a matter of preference.

bubble-wrap-style.yaml
```diff
@@ -3876,6 +3085,29 @@ layers:
...
+                    - kind: entrance
+                all:
+                    - function() { return global.text_visible_address; }
+            draw:
+                mapzen_icon_library:
+                    size: [[13, [95%,95%]], [16, [100%,100%]], [18, [105%,105%]]]
+                    sprite: generic #function() { return feature.kind; }
+                    sprite_default: generic
+                    interactive: global.sdk_interactive
+                    priority: 65 #function() { return (feature.min_zoom && Math.floor(feature.min_zoom * 1000)) || 65; }
```
bcamper: If this is fixed now, should we drop the old function in the comment?

bubble-wrap-style.yaml
```diff
@@ -3876,6 +3085,29 @@ layers:
...
+                    repeat_group: abc
```
bcamper, line 3141: Are we intentionally using separate `repeat_group` logic here, or is this vestigial? Can't remember/tell.

bubble-wrap-style.yaml
```diff
@@ -4700,7 +3936,7 @@ layers:
         draw:
             mapzen_icon_library:
                 visible: global.icon_visible_poi_landuse
-                size: [[13, 14px], [16, 18px], [18, 19px]]
+                size: [[13, [95%,95%]], [16, [100%,100%]], [18, [105%,105%]]]
```
bcamper: Can be simplified to `size: [[13, 95%], [16, 100%], [18, 105%]]`

bubble-wrap-style.yaml
```diff
@@ -5182,7 +4443,7 @@ layers:
             draw:
                 mapzen_icon_library:
                     visible: global.icon_visible_station
-                    size: [[13, 12px], [14, 12px], [15, 16px], [17, 18px], [18, 20px]]
+                    size: [[13, [80%,80%]], [17, [100%,100%]], [18, [110%,110%]]]
```
bcamper: Can be simplified to `size: [[13, 80%], [17, 100%], [18, 110%]]`

bubble-wrap-style.yaml
```diff
@@ -5252,7 +4510,7 @@ layers:
                 kind: [bus_stop]
             draw:
                 mapzen_icon_library:
-                    size: [[18,12px],[20,18px]]
+                    size: [[18, [80%,80%]], [20, [100%,100%]]]
```
bcamper: Can be simplified to `size: [[18, 80%], [20, 100%]]`

bubble-wrap-style.yaml
```diff
@@ -5523,11 +4781,28 @@ layers:
                     visible: true
         military:
             filter:
-                kind: military
+                kind: [military,naval_base]
             draw:
                 dots:
                     color: [0.722,0.769,0.827]
                     visible: true
+        danger_area:
```
bcamper, line 4828: ⚠️⚠️⚠️

bubble-wrap-style.yaml
```diff
@@ -5637,6 +4912,22 @@ layers:
...
+                filter: { kind: [quay,wharf] }
+                draw:
+                    dots:
+                        visible: false
+                    polygons_transparent:
+                        color: [0.8,0.8,0.8,0.5]
+                        visible: true
+                        # hack
+                        order: 236
+            # hack
```
bcamper: What is this hacking?

bubble-wrap-style.yaml
```diff
@@ -5655,8 +4946,54 @@ layers:
...
+                    color: [0.780,0.740,0.538]
+                    visible: true
+
+        plant_nursery:
+            filter:
+                kind: plant_nursery
+            draw:
+                dots:
+                    color: [0.955,0.752,0.650]
+                    visible: true
```
bcamper, line 4997: Are all these `visible: true` lines turning on something that's turned off way up above?

bubble-wrap-style.yaml
```diff
@@ -6122,7 +5476,7 @@ layers:
             draw:
                 mapzen_icon_library:
                     sprite: train_station
-                    size: [[13, 12px], [14, 14px], [15, 16px], [17, 20px]]
+                    size: [[13, [65%,65%]], [16, [100%,100%]]]
```
bcamper: See notes above on how this and other %-based scaling below can be simplified to only use a single value for both dimensions

themes/bubble-wrap-icons-shields-international.yaml
```diff
@@ -0,0 +1,2140 @@
...
+            'IT:SS-1char': [0, 720, 42, 36]
+            'IT:B-road-1char': [0, 720, 42, 36]
+            'ES:A-road-1char': [0, 720, 42, 36]
+            'SE:A-1char': [0, 720, 42, 36]
+            'AT:A-road-1char': [0, 720, 42, 36]
+            'AT:S-road-1char': [0, 720, 42, 36]
+            'TR:highway-1char': [0, 720, 42, 36]
+            'LU:M-1char': [0, 720, 42, 36]
+            'FI:LR-1char': [0, 720, 42, 36]
+
```
bcamper: No need for `all` here, can just be:

`filter: { network: [...], shield_text: true }`

bcamper: No need for `all` here, see comments above.

bcamper: Can be `filter: { network: e-road }` (no array needed)

bcamper: No array needed on filter

bcamper: A lot of the rules that follow only seem to set the `fill` color to white (and/or other small but repeated changes). Can they be consolidated into fewer rules? (Smaller file, faster parsing, etc.)

themes/bubble-wrap-icons-shields-usa.yaml
```diff
@@ -0,0 +1,584 @@
...
+            county_shield-2char: [832, 632, 44, 36]
+            county_shield-3char: [774, 632, 52, 36]
+            county_shield-4char: [700, 632, 68, 36]
+            county_shield-5char: [614, 632, 80, 36]
+            
+layers:
+    roads:
+        shields:
+            shields-usa:
+                filter:
```
bcamper: No `all` needed in filter

themes/bubble-wrap-icons.yaml
```diff
@@ -3,7 +3,7 @@ global:
     # are otherwise defined in the root {style}-style.yaml,
     # themes/color-{theme}.yaml or themes/label-{step} files.
     # This also means {style}-icons.yaml should be imported first.
-    icon_size:          [[13, '14px'], [16, '18px'], [18, '19px']]
+    icon_size:          [[7,[75%,75%]],[12,[100%,100%]],[15,[110%,110%]]]
```
bcamper: Can be `icon_size: [[7, 75%], [12, 100%], [15, 110%]]`

themes/bubble-wrap-icons.yaml
```diff
@@ -42,1652 +49,680 @@ textures:
...
-            'US:NC-3char': [150, 1774, 126, 86]
-            'US:MI-3char': [150, 1774, 126, 86]
-            'US:NC-4char': [0, 1774, 144, 88]
-            'US:MI-4char': [0, 1774, 144, 88]
-            'US:NC-5char': [744, 1682, 168, 86]
-            'US:MI-5char': [744, 1682, 168, 86]
-            'US:ND-1char': [822, 1062, 90, 68]
-            'US:ND-2char': [714, 1062, 102, 68]
-            'US:ND-3char': [592, 1062, 114, 68]
-            'US:ND-4char
```
bcamper, line 762: I need one of these

themes/label-10.yaml
```diff
@@ -12,8 +12,8 @@ global:
     text_visible_island:                true
     icon_visible_landuse_green:         true     # needs work
     text_visible_landuse_green:         true
-    icon_size_green:                    [[13, '14px'], [16, '18px'], [18, '19px']]      # 0px for no, default, more  [[13, '14px'], [16, '18px'], [18, '19px']]
-    icon_size_green_l:                  [[14, '24px'], [16, '32px']]                    # 0px for no, default, more  [[14, '24px'], [16, '32px']]
+    icon_size_green:                    [[13, [95%,95%]], [16, [100%,100%]], [18, [105%,105%]]]
```
bcamper: This and line below can be simplified to have one % value for both dimensions (see similar comments).

themes/label-11.yaml
```diff
@@ -12,8 +12,8 @@ global:
     text_visible_island:                true
     icon_visible_landuse_green:         true     # needs work
     text_visible_landuse_green:         true
-    icon_size_green:                    [[13, '14px'], [16, '18px'], [18, '19px']]      # 0px for no, default, more  [[13, '14px'], [16, '18px'], [18, '19px']]
-    icon_size_green_l:                  [[14, '24px'], [16, '32px']]                    # 0px for no, default, more  [[14, '24px'], [16, '32px']]
+    icon_size_green:                    [[13, [95%,95%]], [16, [100%,100%]], [18, [105%,105%]]]
```
bcamper: Simplify % to single value for both dimensions (see similar comments).

themes/label-6.yaml
```diff
@@ -12,8 +12,8 @@ global:
     text_visible_island:                true
     icon_visible_landuse_green:         true     # needs work
     text_visible_landuse_green:         true
-    icon_size_green:                    [[13, '14px'], [16, '18px'], [18, '19px']]      # 0px for no, default, more  [[13, '14px'], [16, '18px'], [18, '19px']]
-    icon_size_green_l:                  [[14, '24px'], [16, '32px']]                    # 0px for no, default, more  [[14, '24px'], [16, '32px']]
+    icon_size_green:                    [[13, [95%,95%]], [16, [100%,100%]], [18, [105%,105%]]]
```
bcamper: Simplify % to single value for both dimensions (see similar comments).

themes/label-7.yaml
```diff
@@ -12,8 +12,8 @@ global:
     text_visible_island:                true
     icon_visible_landuse_green:         true     # needs work
     text_visible_landuse_green:         true
-    icon_size_green:                    [[13, '14px'], [16, '18px'], [18, '19px']]      # 0px for no, default, more  [[13, '14px'], [16, '18px'], [18, '19px']]
-    icon_size_green_l:                  [[14, '24px'], [16, '32px']]                    # 0px for no, default, more  [[14, '24px'], [16, '32px']]
+    icon_size_green:                    [[13, [95%,95%]], [16, [100%,100%]], [18, [105%,105%]]]
```
bcamper: Simplify % to single value for both dimensions (see similar comments).

themes/label-8.yaml
```diff
@@ -12,8 +12,8 @@ global:
     text_visible_island:                true
     icon_visible_landuse_green:         true     # needs work
     text_visible_landuse_green:         true
-    icon_size_green:                    [[13, '14px'], [16, '18px'], [18, '19px']]      # 0px for no, default, more  [[13, '14px'], [16, '18px'], [18, '19px']]
-    icon_size_green_l:                  [[14, '24px'], [16, '32px']]                    # 0px for no, default, more  [[14, '24px'], [16, '32px']]
+    icon_size_green:                    [[13, [95%,95%]], [16, [100%,100%]], [18, [105%,105%]]]
```
bcamper: Simplify % to single value for both dimensions (see similar comments).

themes/label-9.yaml
```diff
@@ -12,8 +12,8 @@ global:
     text_visible_island:                true
     icon_visible_landuse_green:         true     # needs work
     text_visible_landuse_green:         true
-    icon_size_green:                    [[13, '14px'], [16, '18px'], [18, '19px']]      # 0px for no, default, more  [[13, '14px'], [16, '18px'], [18, '19px']]
-    icon_size_green_l:                  [[14, '24px'], [16, '32px']]                    # 0px for no, default, more  [[14, '24px'], [16, '32px']]
+    icon_size_green:                    [[13, [95%,95%]], [16, [100%,100%]], [18, [105%,105%]]]
```
bcamper: Simplify % to single value for both dimensions (see similar comments).

bcamper (commented):
Left a bunch of comments with some suggested clean-up / simplifications. They may help with maintainability and (a little bit with) performance of scene parsing, but none are blockers.

Huzzah 🎉

themes/bubble-wrap-icons.yaml
nvkelso: ce05472e6916d311fcef4071e6b1f7d53d629295

nvkelso: and d36fedbfef2b5fdf8dadb801f59c8d4bcdfc91ac

themes/label-10.yaml
nvkelso: https://github.com/tangrams/bubble-wrap/commit/ce05472e6916d311fcef4071e6b1f7d53d629295

themes/bubble-wrap-icons-shields-usa.yaml
nvkelso: b38b8910848b62cf3804ad8b70ab04ba8a9d2beb

bubble-wrap-style.yaml
nvkelso: ff50c44c36332a80c4f5b015d167f9a34bf033e3

themes/bubble-wrap-icons-shields-international.yaml
nvkelso: 6bb86d946d44c820b71167801a94575341dbdfea

nvkelso: 336ba5a6078c7b400ae2b061392b0690340cddcf

nvkelso: 6e4ec979ac245de5b815aa25837d382f0571317b

nvkelso: Many of them can, though that makes future disentangling of them for any regionalized style assets more complicated. I'll have a look...

bubble-wrap-style.yaml
nvkelso: I'm setting a useful order to make the features display (otherwise it's 0 and under everything). There is a server issue tracking providing a better value: https://github.com/tilezen/vector-datasource/issues/1588

nvkelso: 63f46138b1a27157a26aa375b2ce71f01dcae99a

bubble-wrap-style.yaml
nvkelso: Yep, to show the dots. Probably not the best structure (for reasons?), but this just extends that.

bubble-wrap-style.yaml
nvkelso: 71f2b36897a07c9749037a7328d3b6c1279a00b6

bubble-wrap-style.yaml
nvkelso: 51f720c3dded7a90fd28492ba28f5bfe52a40d42

bubble-wrap-style.yaml
nvkelso: deeabbec2576afedb8d3c8e28e50e5a1168edef4

bcamper:
The reason the 52f98b7f1c34f85547dcf72efb78b54095d0f717 commit didn't work is that it dropped the `all` while keeping the filters as **arrays**, which have an implicit `any` behavior. An **object** filter has implicit `all` behavior. E.g. see example in 6bb86d946d44c820b71167801a94575341dbdfea:

Should be 
`filter: { network: [...], shield_text: true }`

Or can be written as:
```
filter:
  network: [...]
  shield_text: true
```
But NOT as an array:
`filter: [ network: [...], shield_text: true ]` or
```
filter:
  - network: [...]
  - shield_text: true
```

Does the difference make sense?

nvkelso:
Hmm, it does now but wasn’t obvious before. I prefer to use the explicate
all here as I find it less buggy / confusing.

bcamper:
I definitely find the YAML difference between the "expanded" form (with `-` for arrays, but no preceding character for objects) confusing. For "simple all" filters I like the more compact form `filter: { network: [...], shield_text: true }`, but it is mostly a matter of preference. The extra `all` does have some parsing/memory overhead, but it's likely negligible.

bubble-wrap-style.yaml
nvkelso: It's adding the entrance icons to the same repeat_group as the pois, because they are in different source layers. The only other explicate repeat_group we use is for shields. We set collision false for trees at some zooms.

bubble-wrap-style.yaml
nvkelso: Yep

themes/bubble-wrap-icons.yaml
nvkelso: Lost in spring cleaning...

PR #271, opened 2018-04-24
fix point offset units

tallytalwar:
Refer documentation here: https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/draw.md#offset-points-text

I also tested this on tangram.city/play:

screen shot 2018-04-23 at 6 02 35 pm

screen shot 2018-04-23 at 6 02 23 pm

screen shot 2018-04-23 at 6 02 16 pm

### DIFF
--- a/bubble-wrap-style.yaml
+++ b/bubble-wrap-style.yaml
@@ -3920,7 +3920,7 @@ layers:
                     draw:
                         text-blend-order:
                             text_source: global.ux_language_text_source_boundary_lines
-                            offset: 0px
+                            offset: [0px, 0px]
         other_country_boundary_disputed_etc:
             filter: { kind: [disputed, indefinite, indeterminate, lease_limit, line_of_control, overlay_limit] }
             draw:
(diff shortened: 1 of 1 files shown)


bubble-wrap-style.yaml
```diff
@@ -3906,7 +3906,7 @@ layers:
                         priority: 2
                         visible: global.text_visible_admin
                         text_source: global.ux_language_text_source_boundary_lines_left_right
-                        offset: [0, 5px]
+                        offset: [0px, 5px]
```
matteblair, line 3909: I think omitting units here is ok? There's an example of it in the docs where it describes `offset` with stops.

tallytalwar: Hmm, the documentation mention **no default** for point `offset` units.

tallytalwar: And ES is following the docs and fails when no units are specified for an `offset`.

nvkelso (approved):

PR #274, opened 2018-12-08
Refactor shield functions:

hjanetzek:
Some optimization to reduce the work of js-functios: 
- Use text_source sequence for shield text instead of function.
- To discard shields when text is too long return no sprite instead of using an additional js call to filter them
- Don't allocate memory for building strings
- Don't spend cpu cycles on building strings, hashing and storing them.

### DIFF
--- a/bubble-wrap-style.yaml
+++ b/bubble-wrap-style.yaml
@@ -2371,6 +2371,22 @@ layers:
                         mapzen_icon_library:
                             visible: false
 
+            width_4-5char:
+                filter:
+                    all:
+                        - not:
+                            kind: [minor_road]
+                            network: ['US:I','US:US','US:US:Business','US:US:Truck','US:US:Alternate'] # propably all US:* - bug below?
+                        - |
+                            function() {
+                              var text = feature.shield_text;
+                              var length = text ? text.length : feature.ref.replace(" ", "").length;
+                              return length === 4 || length === 5;
+                            }
+                draw:
+                    mapzen_icon_library:
+                        # this is sensitive to values > 56
+                        priority: 56
             #
             # Now for the good stuff
             #
@@ -2402,6 +2418,7 @@ layers:
                         network: 'US:I'
                     draw:
                         mapzen_icon_library:
+                            priority: 46
                             sprite: |
                                 function() {
                                   switch (feature.shield_text.length) {
@@ -2412,7 +2429,6 @@ layers:
                                     case 5: return 'US:I-5char';
                                   }
                                 }
-                            priority: 46
                 US-I-Business:
                     filter:
                         network: 'US:I:Business'
--- a/bubble-wrap-style.yaml
+++ b/bubble-wrap-style.yaml
@@ -2511,6 +2511,19 @@ layers:
                                     case 5: return 'US:I:Alternate-5char';
                                   }
                                 }
+                # US Federal Routes
+                US-US-common:
+                    filter:
+                        network: ['US:US','US:US:Business','US:US:Truck','US:US:Alternate']
+                        $zoom: { min: 7 }
+                    draw:
+                        mapzen_icon_library:
+                            priority: 47
+                            cull_from_tile: true
+                            text:
+                                font:
+                                    fill: [0.380,0.318,0.312]
+                                    size: [[7,8px],[13,10px],[15,11px]]
                 US-US:
                     filter:
                         network: 'US:US'
@@ -2571,20 +2584,6 @@ layers:
                                     case 5: return 'US:US:Alternate-5char';
                                   }
                                 }
-                # US Federal Routes
-                US-US:
-                    filter:
-                        network: ['US:US','US:US:Business','US:US:Truck','US:US:Alternate']
-                        shield_text: true
-                        $zoom: { min: 7 }
-                    draw:
-                        mapzen_icon_library:
-                            priority: 47
-                            cull_from_tile: true
-                            text:
-                                font:
-                                    fill: [0.380,0.318,0.312]
-                                    size: [[7,8px],[13,10px],[15,11px]]
 
 
         path:
--- a/bubble-wrap-style.yaml
+++ b/bubble-wrap-style.yaml
@@ -2265,8 +2265,8 @@ layers:
                     #       }
                     #       return ['generic_shield-', feature.ref.length, 'char'].join('');
                     #     }
-                    # sprite_default: generic_shield_5char
                     ## [].join('') is good - Better build no new strings at all:
+                    sprite_default: ~ # remove 'generic' default to make discard label-by-sprite function work
                     sprite: |
                         function() {
                           var text = feature.shield_text;
--- a/bubble-wrap-style.yaml
+++ b/bubble-wrap-style.yaml
@@ -2264,7 +2264,7 @@ layers:
                         - [13, 1.50]
                         - [14, 2.0]
                     cull_from_tile: true
-                    visible: false
+                    visible: true
                     text:
                         offset: [0px, -0.5px]
                         repeat_distance: 200px
@@ -2286,7 +2286,6 @@ layers:
                     mapzen_icon_library:
                         priority: 47
                         #color: yellow
-                        visible: global.sdk_road_shields
                 early:
                     filter:
                         - $zoom: { max: 7 }
@@ -2302,7 +2301,6 @@ layers:
                     mapzen_icon_library:
                         priority: 48
                         #color: red
-                        visible: global.sdk_road_shields
                 early:
                     filter:
                         - $zoom: { max: 9 }
@@ -2318,7 +2316,6 @@ layers:
                     mapzen_icon_library:
                         priority: 49
                         #color: aqua
-                        visible: global.sdk_road_shields
                 early:
                     filter:
                         - $zoom: { max: 10 }
@@ -2334,7 +2331,6 @@ layers:
                     mapzen_icon_library:
                         priority: 53
                         #color: blue
-                        visible: global.sdk_road_shields
                 early:
                     filter:
                         - $zoom: { max: 12 }
@@ -2350,7 +2346,6 @@ layers:
                     mapzen_icon_library:
                         priority: 55
                         #color: green
-                        visible: global.sdk_road_shields
                 early:
                     filter:
                         - $zoom: { max: 13 }
@@ -2363,7 +2358,6 @@ layers:
                 draw:
                     mapzen_icon_library:
                         priority: 56
-                        visible: global.sdk_road_shields
                 early:
                     filter:
                         - $zoom: { max: 14 }
--- a/bubble-wrap-style.yaml
+++ b/bubble-wrap-style.yaml
@@ -2191,6 +2191,7 @@ layers:
                             stroke: { color: [1.00,1.00,1.00], width: 2 }
 
         shields:
+            enabled: global.sdk_road_shields
             filter:
                 all:
                     - $zoom: { min: 6 }
@@ -2204,7 +2205,6 @@ layers:
                     # that makes United States look a bit funky with `(xxx)` and generic shields
                     # showing up in fields of custom state shields like US:CA and US:NY
                     #
-                    - function() { return global.sdk_road_shields; }
                     - any:
                         - shield_text: true    # some roads don't have shield text, deal with that later
                         - all:
@@ -2215,15 +2215,40 @@ layers:
             draw:
                 mapzen_icon_library:
                     # you need to match any custom shield to the vector tile `network` values
+                    # sprite: |
+                    #     function() {
+                    #       if( feature.shield_text ) {
+                    #         return ['generic_shield-', feature.shield_text.length, 'char'].join('');
+                    #       }
+                    #       return ['generic_shield-', feature.ref.length, 'char'].join('');
+                    #     }
+                    # sprite_default: generic_shield_5char
+                    ## [].join('') is good - Better build no new strings at all:
                     sprite: |
                         function() {
-                            if( feature.shield_text ) {
-                                return ('generic_shield-' + feature.shield_text.length + 'char');
-                            } else {
-                                return ('generic_shield-' + feature.ref.length + 'char');
+                          var text = feature.shield_text;
+                          if (text === undefined) {
+                            text = feature.ref;
+                            if (text === undefined) { return undefined; }
+                            if (text.length > 3) {
+                              text = text.replace(" ", "");
                             }
+                          }
+                          switch (text.length) {
+                            case 1:
+                                return 'generic_shield-1char';
+                            case 2:
+                                return 'generic_shield-2char';
+                            case 3:
+                                return 'generic_shield-3char';
+                            case 4:
+                                if ($zoom < 11) break;
+                                return 'generic_shield-4char';
+                            case 5:
+                                if ($zoom < 15) break;
+                                return 'generic_shield-5char';
+                          }
                         }
-                    sprite_default: generic_shield_5char
                     size: [[7,[75%,75%]],[12,[100%,100%]],[15,[110%,110%]]]
                     priority: 58
                     # Size is specified below, and limited to 5 char in another filter
@@ -2244,14 +2269,7 @@ layers:
                         offset: [0px, -0.5px]
                         repeat_distance: 200px
                         anchor: center
-                        text_source: |
-                            function() {
-                                if( feature.shield_text ) {
-                                    return feature.shield_text;
-                                } else {
-                                    return feature.ref;
-                                }
-                            }
+                        text_source: [shield_text, ref]
                         font:
                             family: Montserrat
                             weight: 500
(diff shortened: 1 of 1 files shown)


hjanetzek:
For our test-tile this change reduces the number of js calls to 3733 filter +  24930 style calls from previously 12633 filter + 27087 style function calls. 
https://gist.github.com/hjanetzek/addc9fd34a1d99e0621c07fb43bf48c0
https://gist.github.com/hjanetzek/07efe63492680f9f3831832966aa1dae
And reduces js evaluation time on my duktape-optimizaiton branch from 47ms to 31ms

bubble-wrap-style.yaml
```diff
@@ -78,68 +116,73 @@ global:
...
-            /*
-            var name = (global.ux_language && feature['name:'+global.ux_language]) || (global.ux_language_fallback && feature['name:'+global.ux_language_fallback]) || feature['name'];
-            if (feature.ref && (feature.ref.length < 6) && name) {
-                return feature.ref + ' ' + name;
-            } else {
-                return name;
-            }
-            */
-        }
+    ux_language_text_source_road_ref_and_name: global.ux_language_text_source
```
bcamper, line 148: Definitely better to just reuse the existing function like this where possible, but this also reminded me, I added a simple cache to JS recently to reuse scene JS functions internally when they have the same source (similar to caching shader programs by source), and it can help catch cases like this.

bubble-wrap-style.yaml
```diff
@@ -2239,19 +2307,12 @@ layers:
...
                         anchor: center
-                        text_source: |
-                            function() {
-                                if( feature.shield_text ) {
-                                    return feature.shield_text;
-                                } else {
-                                    return feature.ref;
-                                }
-                            }
+                        text_source: [shield_text, ref]
```
bcamper, line 2315: Native scene syntax FTW!!! 👌

bubble-wrap-style.yaml
```diff
@@ -2446,37 +2436,154 @@ layers:
                         - shield_text: true
                 draw:
                     mapzen_icon_library:
-                        #texture: mapzen_icon_library_shields_usa
+                        # texture: mapzen_icon_library_shields_usa
                         # you need to match any custom shield to the vector tile `network` values
-                        sprite: function() { return (feature.network + '-' + feature.shield_text.length + 'char'); }
-                        # FYI: sprite_default doesn't support functions, default is carried by parent style's sprite function
+                        sprite_default: |
```
bcamper, line 2441: Note `sprite_default` doesn't support functions, I think this logic was moved up to the top of the shield layer now?

bubble-wrap-style.yaml
```diff
@@ -2446,37 +2436,154 @@ layers:
...
-                        shield_text: true
                     draw:
                         mapzen_icon_library:
                             priority: 46
-                            # always show (or not show), irrespective of zoom
-                            visible: global.sdk_road_shields
+                            sprite: |
+                                function() {
+                                  switch (feature.shield_text.length) {
+                                    case 1: return 'US:I-1char';
```
bcamper, line 2462: This switch-case style is great for these.

bcamper:
I wasn't able to discern difference in performance loading this version in Tangram JS (but hey, it didn't get any worse either :)

bubble-wrap-style.yaml
```diff
@@ -14,57 +14,95 @@ global:
     #ux/ui
     ux_language: false             # l10n language code, trusting OSM in v0.10 tiles, fixed in v1.0 tiles
     ux_language_fallback: false    # l10n language code, trusting OSM in v0.10 tiles, fixed in v1.0 tiles
+    # if a ux_langauge has been defined use that, else if there is feature name
```
nvkelso, line 17: Nit: add line space above here as it's getting a little crowded to read now.

bubble-wrap-style.yaml
```diff
@@ -78,68 +116,73 @@ global:
...
-            var name = (global.ux_language && feature['name:'+global.ux_language]) || (global.ux_language_fallback && feature['name:'+global.ux_language_fallback]) || feature['name'];
-            if (feature.ref && (feature.ref.length < 6) && name) {
-                return feature.ref + ' ' + name;
-            } else {
-                return name;
-            }
-            */
-        }
+    ux_language_text_source_road_ref_and_name: global.ux_language_text_source
+    # ux_language_text_source_road_ref_and_name: |
```
nvkelso, line 149: I'm okay removing the commented out `ux_language_text_source_road_ref_and_name` section – it's vestigial from when Tangram didn't really support road shields.

bubble-wrap-style.yaml
```diff
@@ -78,68 +116,73 @@ global:
...
+    #         var name = (global.ux_language && feature['name:'+global.ux_language]) ||
+    #                    (global.ux_language_fallback && feature['name:'+global.ux_language_fallback]) || feature['name'];
+    #         if(feature.ref && name) {
+    #             return (feature.ref + ' ' + name);
+    #         } else {
+    #             return name;
+    #         }
+    #     }
+    ux_language_text_source_road_ref_and_name_short: global.ux_language_text_source
+    # ux_language_text_source_road_ref_and_name_short: |
```
nvkelso, line 160: I'm okay removing the commented out `ux_language_text_source_road_ref_and_name_short` section – it's vestigial from when Tangram didn't really support road shields.

bubble-wrap-style.yaml
```diff
@@ -14,57 +14,95 @@ global:
...
+    # ux_language_text_source_boundary_lines: |
+    #    function() {
+    #        var right = (global.ux_language && feature['name:right:'+global.ux_language]) ||
+    #                    (global.ux_language_fallback && feature['name:right:'+global.ux_language_fallback]) ||
+    #                    feature['name:right'];
+    #        var left = (global.ux_language && feature['name:left:'+global.ux_language]) ||
+    #                   (global.ux_language_fallback && feature['name:left:'+global.ux_language_fallback]) ||
+    #                   feature['name:left'];
+    #        if( right && left ) {
+    #            if( right.includes(' ') || left.includes(' ') ) {
```
nvkelso, line 65: Removing this if/else logic for stacking or not stacking the resulting `ux_language_text_source_boundary_lines ` text string would be a visual regression.

bubble-wrap-style.yaml
nvkelso: Seems worth a separate Tangram ES issue... @hjanetzek @matteblair?

bubble-wrap-style.yaml
```diff
@@ -78,68 +116,73 @@ global:
...
+    #             return feature.ref + ' ' + name;
+    #         } else {
+    #             return name;
+    #         }
+    #     }
     ux_language_text_source_piste_advanced: |
         function() {
-            var name = (global.ux_language && feature['name:'+global.ux_language]) || (global.ux_language_fallback && feature['name:'+global.ux_language_fallback]) || feature['name'];
+            var name = (global.ux_language && feature['name:'+global.ux_language]) ||
+                       (global.ux_language_fallback && feature['name:'+global.ux_language_fallback]) || feature['name'];
```
nvkelso, line 173: nit: The convention above in this PR is to separate each || onto a new line, I think `feature['name'];` here as well?

bubble-wrap-style.yaml
```diff
@@ -78,68 +116,73 @@ global:
...
             return name ? ('◆ ' + name) : '◆';
         }
     ux_language_text_source_piste_expert: |
         function() {
-            var name = (global.ux_language && feature['name:'+global.ux_language]) || (global.ux_language_fallback && feature['name:'+global.ux_language_fallback]) || feature['name'];
+            var name = (global.ux_language && feature['name:'+global.ux_language]) ||
+                       (global.ux_language_fallback && feature['name:'+global.ux_language_fallback]) || feature['name'];
```
nvkelso, line 179: nit: The convention above in this PR is to separate each || onto a new line, I think `feature['name'];` here as well?

bubble-wrap-style.yaml
```diff
@@ -78,68 +116,73 @@ global:
...
             return name ? ('◆◆ ' + name) : '◆◆';
         }
     ux_language_text_source_building_and_address: |
         function() {
-            var name = (global.ux_language && feature['name:'+global.ux_language]) || (global.ux_language_fallback && feature['name:'+global.ux_language_fallback]) || feature['name'];
+            var name = (global.ux_language && feature['name:'+global.ux_language]) ||
+                       (global.ux_language_fallback && feature['name:'+global.ux_language_fallback]) || feature['name'];
```
nvkelso, line 185: nit: The convention above in this PR is to separate each || onto a new line, I think `feature['name'];` here as well?

bubble-wrap-style.yaml
```diff
@@ -2191,6 +2234,7 @@ layers:
                             stroke: { color: [1.00,1.00,1.00], width: 2 }
 
         shields:
+            enabled: global.sdk_road_shields
```
nvkelso, line 2237: Logically this makes sense, but I recall "reasons" for why that was backed out to be a filter instead. I'll need to review the old PRs...

bubble-wrap-style.yaml
nvkelso: So, it turns out `shield_text` != `ref`, and more specifically `shield_text.length()` != `ref.length()`.

It looks like you've accounted for that in the new sprite function (above), though, so this logic would always be coherent (modulo my 1 comment about stopping stripping spaces from ref above).

bubble-wrap-style.yaml
```diff
@@ -2215,15 +2258,40 @@ layers:
...
                         function() {
-                            if( feature.shield_text ) {
-                                return ('generic_shield-' + feature.shield_text.length + 'char');
-                            } else {
-                                return ('generic_shield-' + feature.ref.length + 'char');
+                          var text = feature.shield_text;
+                          if (text === undefined) {
+                            text = feature.ref;
+                            if (text === undefined) { return undefined; }
+                            if (text.length > 3) {
```
nvkelso, line 2276: This should only be done if the text is coming from `ref` not `shield_text`, as `shield_text` deliberately sometimes keeps whitespace in the value. (So this logic is good.)

But later on below... since you propose to set `text_source: [sheild_text, ref]`, the symbol size will no longer match after you modify the interpreted value of **ref** here. So we either need to stop stripping the space from `ref` **here**, or add the stripping to the text source below. I'd be fine stopping stripping of the space from ref here as the vector data is soooo much better starting in Tilezen v1.5 tiles.

bubble-wrap-style.yaml
nvkelso: Yeah, this section doesn't look right. Shouldn't it still have the `sprite: ` function included?

bubble-wrap-style.yaml
nvkelso: Hmm. That's may be fine to expand just for this USA example here with switch statements... but seems rather excessive to bloat the full USA and international shield theme imports. Those could easily get 5x as large as they are today in terms of line count. It's not impossible since much of that is auto-generated, but my goodness!

If the switch is more performant... could we say instead:

```
                                case 1: return feature.network + '-1char';
                                case 2: return feature.network + '-2char';
                                case 3: return feature.network + '-3char';
                                case 4: return feature.network + '-4char';
                                case 5: return feature.network + '-5char';
```

bubble-wrap-style.yaml
```diff
@@ -2345,99 +2401,33 @@ layers:
...
-                            return (feature.shield_text.length === 4)
-                        } else {
-                            return (feature.ref.replace(" ", "").length === 4)
-                        }
-                    }
-                draw:
-                    mapzen_icon_library:
-                        # this is sensitive to values > 56
-                        priority: 56
-                early:
```
nvkelso, line 2396: This moved to:

- https://github.com/tangrams/bubble-wrap/pull/274/files#diff-d5306f7f58640a806f55ed4eda76d550R2288

bubble-wrap-style.yaml
```diff
@@ -2345,99 +2401,33 @@ layers:
...
+                            function() {
+                              var text = feature.shield_text;
+                              var length = text ? text.length : feature.ref.replace(" ", "").length;
+                              return length === 4 || length === 5;
+                            }
                 draw:
                     mapzen_icon_library:
                         # this is sensitive to values > 56
                         priority: 56
-                early:
```
nvkelso, line 2414: This moved to: 

- https://github.com/tangrams/bubble-wrap/pull/274/files#diff-d5306f7f58640a806f55ed4eda76d550R2291

bubble-wrap-style.yaml
```diff
@@ -2215,15 +2258,40 @@ layers:
             draw:
                 mapzen_icon_library:
                     # you need to match any custom shield to the vector tile `network` values
+                    # sprite: |
```
nvkelso, line 2261: Can just remove this commented out section?

bubble-wrap-style.yaml
```diff
@@ -2215,15 +2258,40 @@ layers:
...
+                    #     function() {
+                    #       if( feature.shield_text ) {
+                    #         return ['generic_shield-', feature.shield_text.length, 'char'].join('');
+                    #       }
+                    #       return ['generic_shield-', feature.ref.length, 'char'].join('');
+                    #     }
+                    ## [].join('') is good - Better build no new strings at all:
+                    sprite_default: ~ # remove 'generic' default to make discard label-by-sprite function work
```
nvkelso, line 2269: Don't need this about `sprite_default` at all now?

(Or at least I don't understand what the ~ does in YAML / scene files.)

bubble-wrap-style.yaml
```diff
@@ -2239,19 +2307,12 @@ layers:
                         - [13, 1.50]
                         - [14, 2.0]
                     cull_from_tile: true
-                    visible: false
+                    visible: true
```
nvkelso, line 2310: Turning this on by default in the past has caused _too_ many shields to show up. We've cleaned up shields server side quite a bit since then, so worth reevaluating. But this may cause visual problems.

bubble-wrap-style.yaml
```diff
@@ -2332,7 +2389,6 @@ layers:
                     mapzen_icon_library:
                         priority: 55
                         #color: green
```
nvkelso, line 2391: Might as well take this debug line out, too? ;)

bubble-wrap-style.yaml
```diff
@@ -2345,99 +2401,33 @@ layers:
...
                     mapzen_icon_library:
                         # this is sensitive to values > 56
                         priority: 56
-                early:
-                   filter: { $zoom: { max: 13 } }
-                   draw:
-                       mapzen_icon_library:
-                           visible: false
-
-            # this is kinda a hack – generally we don't have artwork sized 6+
```
nvkelso, line 2420: This is implied now – in that no sprite is defined for cases where the size is not in the range 1-5.

bubble-wrap-style.yaml
```diff
@@ -2345,99 +2401,33 @@ layers:
...
-                        } else {
-                            return (feature.ref.replace(" ", "").length === 5)
-                        }
-                    }
+            width_4-5char:
+                filter:
+                    all:
+                        - not:
+                            kind: [minor_road]
+                            network: ['US:I','US:US','US:US:Business','US:US:Truck','US:US:Alternate'] # propably all US:* - bug below?
```
nvkelso, line 2416: I think the old logic's purpose was to just change the priority, not limit it to specific kinds or networks. So remove these lines but keep the function on the text length below?

bubble-wrap-style.yaml
```diff
@@ -2446,37 +2436,154 @@ layers:
                         - shield_text: true
                 draw:
                     mapzen_icon_library:
-                        #texture: mapzen_icon_library_shields_usa
+                        # texture: mapzen_icon_library_shields_usa
```
nvkelso, line 2439: This line should just be removed.

bubble-wrap-style.yaml
```diff
@@ -2215,15 +2258,40 @@ layers:
...
+                          }
+                          switch (text.length) {
+                            case 1:
+                                return 'generic_shield-1char';
+                            case 2:
+                                return 'generic_shield-2char';
+                            case 3:
+                                return 'generic_shield-3char';
+                            case 4:
+                                if ($zoom < 11) break;
```
nvkelso, line 2288: 👍

bubble-wrap-style.yaml
nvkelso: (It might be related to a fix in Tangram v0.16 for reevaluating globals?)

nvkelso:
(I haven't tested this visually yet, but will after some of the above are addressed.)

bubble-wrap-style.yaml
hjanetzek: We compile identical functions only one time (per js context) as well :)

bubble-wrap-style.yaml
hjanetzek: This would still create new strings for each feature in js context which is quite expensive (at least it is in Duktape where all strings are interned, meaning the new string is hashed and then checked for identical strings to have only one instance in memory) 
What I thought of was that we could provide a stringbuilder function in js so that we could have a more efficient native implementation. This would require to return feature properties as proxy string objects instead of js strings from the getter of our feature proxy object, etc implementation details...

On the js side it could look like:`return stringbuilder(feature.network,'-1char');`
 
 Edit: Oh - the string proxy would be an additional optimization to not have to copy into and intern 'feaure.network' in js context. The stringbuilder could be done without it.

bcamper:
I'd like to avoid any platform-specific implementation concepts leaking
into the scene file like that.

hjanetzek:
@bcamper What I meant was that Tangram can provide these helper functions for all implementations. Then we can recommend using `stringbuilder()` over `+` or `[].join("")`

bcamper:
Yes I get it, but that seems very awkward to me.

hjanetzek:
There seem to be some stringbuilder implementation for js to avoid the overhead of the basic functions as well.

hjanetzek:
Though it's certainly best to have no string building at all

bcamper:
@hjanetzek I think I'm not following what you mean here, can you clarify with examples?

> There seem to be some stringbuilder implementation for js to avoid the overhead of the basic functions as well.

If you suggest introducing a custom string function into the user-defined scene JS functions, that seems 1) like an non-ideal and not obvious exposure of internal implementation details to the user and 2) is motivated not just by ES but by implementation characteristics of the Duktape engine specifically (if I understand correctly). Further, implementing a no-op function wrapper like that for compatibility in Tangram JS could even have negative performance implications by adding another function call (though I suspect it wouldn't be significant in this case, it's another example of why I want to avoid tailoring like that in principle). Apologies if I'm misinterpreting your proposal though.

I just think we should look for more generic solutions in such cases. For example, since this kind of string substitution is pretty common, we could have dedicated syntax for it (e.g. following our tile URL template syntax):

`sprite: "US:I-{feature.shield_text.length}char"`
(note double-quotes are just because of `:` inside YAML value)

Then we've got shorter scene syntax and more native control over how it's implemented. This follows the principle that the scene JS functions are ideal for truly custom logic, and for prototyping new features, but ultimately, we can provide native scene syntax for "core"/common needs. Note, I'm not saying we actually *should* add such syntax right now, without better context of the various tradeoffs in speed and file size (IMO the switch you proposed could be just fine, or maybe the prior dynamic version is) :)

hjanetzek:
Ah right, a native syntax would be even better -  I got distracted from thinking about this approach by too much Duktape optimization work..

PR #275, opened 2019-02-15
True, not yes

bcamper:
It's `visible: true`, not `visible: yes` :)


nvkelso (approved):
Gah, I thought we had caught all those `yes` before! Thanks :)

PR #277, opened 2019-07-18
add POV and other v1.8, v1.7 changes

nvkelso:
- ⭐️Add POV for boundaries and capitals; add map unit lines (eg Wales and Scotland)
    - tilezen_v1d8_disputed_boundaries
- 🚚Add heavy goods vehicle (hgv) overlay
- 🤑Add toll road overlay
- 🚧 Add construction roads
- 🏞 Add new landuse treatments
- 👨 Fix `aboriginal_lands` boundary line treatment
- 🔋 Fix `generator` icons to use kind_detail lookup
- 🐯 Fix zoo `enclosure` labels
- 🚘 Fix motorway junction (exit) labels
- ☠️ Fix `danger_area` and `range` landuse styling (not RED!)
- 🏷 Update `labels-11` theme to not reference all the client side hacks that were fixed server side
- 💻 Remove bunch of logic that moved server side (like for filtering out bad features, setting label collision priority)
- 🎉 Upgrade Tangram JS to v0.19


bubble-wrap-style.yaml
```diff
@@ -12,8 +12,96 @@ global:
...
+                    return true;
+                case( 'unrecognized_country' ):
+                    return true;
+                case( 'unrecognized_region' ):
+                    return true;
+                default:
+                    return false;
+            }
+        }
+    ux_point_of_view_kind_filter_country_capital: |
```
nvkelso, line 83: @bcamper Can you take a look at this section? When viewing Taipei the capital icon should turn OFF when in the Chinese POV as `country_capital:cn` is **false** (even though the default `country_capital` is **true**). The same pattern we used for the name fallbacks doesn't work here.

bubble-wrap-style.yaml
```diff
@@ -1032,34 +1141,18 @@ layers:
...
                 kind: ocean
             draw:
                 grid:
                     color: global.water1
+                    
         inland-water:
             filter:
                 all:
                     - kind: [lake, water, riverbank, reservoir, swimming_pool]
-                any:
```
nvkelso, line 1045: This is all taken care of server side now.

bubble-wrap-style.yaml
```diff
@@ -1175,39 +1273,18 @@ layers:
                         - $geometry: point
                         - not: { kind: [riverbank, dock, playa] }
                         - area: true
-                    any:
```
nvkelso, line 1178: all water filtering is server side now

bubble-wrap-style.yaml
```diff
@@ -1118,18 +1211,23 @@ layers:
                     lines:
                         color: [[8,[0.667,0.792,0.824]],[11,[0.698,0.808,0.835]],[13,[0.710,0.796,0.824]],[14,[0.522,0.792,0.855]]]
                         width: [[9,0.5px],[10,0.75px],[11,0.75px],[12,0.8px],[13,1px],[14,1px],[15,1px],[16,1px],[17,1.5px]]
+#            intermittent:
```
nvkelso: Not sure this should be removed... but it wasn't active here...

bubble-wrap-style.yaml
```diff
@@ -1175,39 +1273,18 @@ layers:
...
                     draw:
                         text-blend-order:
                             text_source: global.ux_language_text_source
+                            text_wrap: 8
                             font:
                                 fill: [0.314,0.591,0.640] #global.text_fill_water
                                 family: global.text_font_family
                                 style: italic
                                 weight: normal
-                                size: 12px
```
nvkelso, line 1203: This is fixed server side now.

bubble-wrap-style.yaml
```diff
@@ -1263,20 +1340,6 @@ layers:
                         - $geometry: point
                         - kind: playa
                         - area: true
-                    any:
```
nvkelso, line 1266: Moved server side.

bubble-wrap-style.yaml
```diff
@@ -1382,11 +1446,96 @@ layers:
             text-blend-order:
                 text_source: global.ux_language_text_source
                 visible: false    # labels are enabled by each layer below
+                priority: |
+                    function() { return (feature['min_zoom'] + (1 - 1 / feature['collision_rank']) * 0.1) || 65; }
```
nvkelso: Use the v1.7 `collision_rank` for predictable POI labeling. 

Really we should also have a server budget and do some collisions there for low- and mid-zooms.

bubble-wrap-style.yaml
```diff
@@ -1382,11 +1446,96 @@ layers:
...
+#                          var color = '#';
+#                          for (var i = 0; i < 6; i++ ) {
+#                              color += letters[Math.floor(Math.random() * 16)];
+#                          }
+#                          return color;
+#                      }
+#                   order: 1000
+#                   width: 1px
+#                   visible: true
+#        z-hgv:
```
nvkelso: This should be turned into a new overlay in the DAT gui.

bubble-wrap-style.yaml
```diff
@@ -1493,7 +1642,7 @@ layers:
                     $zoom: [13,14]
                 draw:
                     text-blend-order:
-                        priority: 50
+#                        priority: 50
```
nvkelso: These should just be removed... they are all set above now from the data.

bubble-wrap-style.yaml
```diff
@@ -3154,13 +3317,16 @@ layers:
...
+                #color: red
+                #width: [[9, 2px], [14, 2px], [16, 3px], [17, 8m]]
         country:
             filter:
                 any:
-                    - kind: country
-                    - admin_level: 2
+                    #- kind: [country, indefinite, indeterminate]
+                    #- admin_level: 2
+                    - global.ux_point_of_view_kind_filter_country
```
nvkelso, line 3339: Added new POV filter for v1.7 and v1.8

bubble-wrap-style.yaml
```diff
@@ -3176,7 +3342,9 @@ layers:
                 filter: { name: true, $zoom: { min: 8 } }
                 draw:
                     text-blend-order:
-                        priority: 2
+#                        priority: 2
```
nvkelso: Remove this line, rely on the next line for server priority

bubble-wrap-style.yaml
```diff
@@ -3194,27 +3362,79 @@ layers:
                         text-blend-order:
                             text_source: global.ux_language_text_source_boundary_lines
                             offset: [0px, 0px]
+
         other_country_boundary_disputed_etc:
-            filter: { kind: [disputed, indefinite, indeterminate, lease_limit, line_of_control, overlay_limit] }
-            draw:
+            filter: 
+                global.ux_point_of_view_kind_filter_disputed_etc
+                #- kind: [disputed, lease_limit, line_of_control, overlay_limit]
```
nvkelso: Can remove these 2 lines, they are carried in the global now

bubble-wrap-style.yaml
```diff
@@ -3194,27 +3362,79 @@ layers:
...
+                #- kind: [disputed_breakaway, disputed_claim, disputed_elusive, disputed_reference_line]
+            draw:
+                #lines:
```
nvkelso: This lines style were just for debug, remove them

bubble-wrap-style.yaml
```diff
@@ -3194,27 +3362,79 @@ layers:
...
+                    visible: true
                     order: global.feature_order
                     color: global.country_boundary
-                    width: [[1, 0.2px], [2, 0.5px], [4, 1.0px], [9, 1.5px], [14, 2.5px], [16, 3.5px], [17, 10m]]
+                    width: [[1, 0.2px], [2, 0.5px], [5, 1.5px], [6, 2.0px], [8, 2.5px], [16, 3.5px], [17, 10m]]
             disputed:
-                filter: { kind: [disputed, line_of_control] }
-                draw:
+                filter: 
+                    #- kind: [disputed, line_of_control]
```
nvkelso: Remove these lines, they are carried in the global now.

bubble-wrap-style.yaml
```diff
@@ -3194,27 +3362,79 @@ layers:
...
+                    #- kind: [disputed_breakaway, disputed_claim, disputed_elusive, disputed_reference_line]
+                    global.ux_point_of_view_kind_filter_disputed_star
+                draw:
+                    #lines:
```
nvkelso: Remove this lines debug style

bubble-wrap-style.yaml
```diff
@@ -3194,27 +3362,79 @@ layers:
...
                     dashed:
                         visible: false
                     dashed_big:
+                        visible: true
                         order: global.feature_order
                         color: global.country_boundary
                         #width: [[1, 0.2px], [2, 0.5px], [4, 1.0px], [9, 1.5px], [14, 2.5px], [16, 3.5px], [17, 10m]]
                         width: [[1, 0.5px], [2, 1.0px], [4, 1.5px], [9, 2.5px], [14, 3.5px], [16, 4.5px], [17, 14m]]
+                        #width: 20px
+#            disputed_star_early:
```
nvkelso: This seems like it's junk and should be removed?

bubble-wrap-style.yaml
```diff
@@ -3194,27 +3362,79 @@ layers:
...
+#                    #kind: [disputed_breakaway, disputed_claim, disputed_elusive, disputed_reference_line]
+#                    global.ux_point_of_view_kind_filter_disputed_star
+#                    $zoom: [5,6]
+#                draw:
+#                    dashed:
+#                        width: [[4, 0px], [5, 1.0px], [8, 2.5px], [16, 3.5px], [17, 10m]]
+#                    dashed_big:
+#                        width: [[5, 0px], [5, 1.0px], [9, 2.5px], [14, 3.5px], [16, 4.5px], [17, 14m]]
+
+        unrecognized:
```
nvkelso, line 3395: New for POV boundaries

bubble-wrap-style.yaml
```diff
@@ -3194,27 +3362,79 @@ layers:
...
+                 lines:
+                     visible: false
+                 text-blend-order:
+                     visible: false
+                 dashed:
+                     visible: false
+                 dashed_big:
+                     visible: false
+
+        map_unit:
```
nvkelso, line 3408: Added a new kind (like for England, Wales)

bubble-wrap-style.yaml
```diff
@@ -3232,11 +3452,21 @@ layers:
                 draw:
                     lines:
                         visible: false
+
+            aboriginal_lands:
```
nvkelso, line 3438: EG around 10/41.1154/-483.5195

bubble-wrap-style.yaml
```diff
@@ -3285,7 +3538,7 @@ layers:
                 $zoom: { min: 2, max: 9 }
             draw:
                 text-blend-order:
-                    priority: 3
+#                    priority: 3
```
nvkelso: These old priority should be removed.

bubble-wrap-style.yaml
```diff
@@ -3920,7 +4225,7 @@ layers:
                     # there are more labels than we want to show, so we hide them,
                     # then selectively reveal in classes below (ultimately a data problem)
                     visible: global.text_visible_neighbourhoods_e
-                    priority: 29
+#                    priority: 29
```
nvkelso: Remove old priority

bubble-wrap-style.yaml
```diff
@@ -4042,291 +4357,75 @@ layers:
...
-                    mapzen_icon_library:
-                        visible: false
-            hide-until-z19-any:
-                filter:
-                    $zoom: { max: 19 }
-                    kind: [bicycle_parking, burial_vault, car_sharing]
-                draw:
-                    mapzen_icon_library:
-                        visible: false
+#        no-name:
```
nvkelso: This section should be removed, it's all server side now and working.

bubble-wrap-style.yaml
```diff
@@ -4042,291 +4357,75 @@ layers:
...
-                    mapzen_icon_library:
-                        visible: false
 
         # set special priority, sprite, and text styles per kind
-        has-area:
-            filter: { area: true }
-            draw:
-                mapzen_icon_library:
-                    priority: 41
+#        has-area:
```
nvkelso: Should also be removed.

bubble-wrap-style.yaml
```diff
@@ -4457,19 +4558,19 @@ layers:
...
                             style: italic
                             fill: '#4ba3c5'
 
-        landuse-labels-not-any-above:
-            filter:
-                not: { kind: [aerodrome, airport, allotments, aquarium, attraction, battlefield, beach, cemetery, college, conservation, forest, garden, glacier, golf_course, grass, kindergarten, landmark, maze, national_park, nature_reserve, park, playground, protected_area, resort, school, stadium, station, theme_park, university, winery, winter_sports, zoo] }
-            draw:
-                mapzen_icon_library:
-                    priority: 43
+#        landuse-labels-not-any-above:
```
nvkelso: Remove

bubble-wrap-style.yaml
```diff
@@ -4652,45 +4758,45 @@ layers:
...
-#                           } else if( feature.scale_rank == 1 ) {
-#                               return [0.,0.5,0.,1];
-#                           } else if( feature.scale_rank == 0 ) {
-#                               return [0.5,0.,0.,1];
-#                           } else {
-#                               return [0.5,0.5,0.5,1];
-#                           }
-#                       }
-#                   }
+#                #debug for polygon merging
```
nvkelso, line 4568: This building section needs revisiting... I had a bad merge conflict along the way.

bubble-wrap-style.yaml
```diff
@@ -4959,13 +5075,13 @@ layers:
...
                         # hack the order as the server isn't provided a good value
-                        order: 236
+                        #order: 236
             # hack the order as the server isn't provided a good value
-            shipyard:
-                filter: { kind: [shipyard,container_terminal,ferry_terminal,port_terminal] }
-                draw:
-                    dots:
-                        order: 237
+            #shipyard:
```
nvkelso: Remove this since fixed on server

bubble-wrap-style.yaml
```diff
@@ -5071,6 +5185,56 @@ layers:
                 draw:
                     dots:
                         color: [0.584,0.843,0.792]
+        grass:
```
nvkelso, line 4989: New landcover colors for new data

bubble-wrap-style.yaml
bcamper: Believe this is fixed in e5e6f31075230c78fff1d2acf53bd064fa8811b8. The issue is that we don't want the *absence* of a `capital_country:XX` value to turn a capital *off*. We want to check for `!= null` so that we only consider values that are explicitly `true` or `false`.

bubble-wrap-style.yaml
```diff
@@ -655,6 +744,16 @@ styles:
                     color.rgb = mix(color.rgb, vec3(0.850), TileDots(45., 0.1));
                     // size was set to 35 previously, now 45
 
+    dots_transparent:
```
nvkelso: Removed this unused section in 942074aacc99e58a82c8dfc080f4c4125a8093fc.

bubble-wrap-style.yaml
nvkelso: Restored in 84584c5b55e8c4126ab45ac39c8d15ecc821380b.

bubble-wrap-style.yaml
nvkelso: Fixed in 0c2624e3cb61fceb0555934313ffbccd251a0be5.

bubble-wrap-style.yaml
nvkelso: Fixed via 8cbbc7f15e702d7df72ecfcfd0df8e32d170e4fa.

bubble-wrap-style.yaml
nvkelso: Fixed in 0fb72303a983c8c27d45fb2c7be9a941e0138dc3.

bubble-wrap-style.yaml
nvkelso: Fixed in be9020163d755c795db104b0fb05bf5d4144d44d.

bubble-wrap-style.yaml
nvkelso: Seems fine, actually.

nvkelso:
@bcamper This PR is cleaned up and ready for new 👀, please!

bubble-wrap-style.yaml
bcamper: 🤘🤖

bubble-wrap-style.yaml
```diff
@@ -1175,39 +1259,18 @@ layers:
...
-                        - { $zoom: { min: 10 }, area: { min: 100000000 } }
-                        - { $zoom: { min: 11 }, area: { min: 10000000 } }
-                        - { $zoom: { min: 12 }, area: { min: 2000000 } }
-                        - { $zoom: { min: 13 }, area: { min: 1000000 } }
-                        - { $zoom: { min: 14 }, area: { min: 500000 } }
-                        - { $zoom: { min: 15 }, area: { min: 100000 } }
-                        - { $zoom: { min: 15 }, area: { min: 50000 } }
-                        - { $zoom: { min: 16 }, area: { min: 20000 } }
-                        - { $zoom: { min: 17 } }
+                        - not: { kind: [sea, ocean] }
```
bcamper, line 1320: Can't this be merged into the `not: { kind: [...] }` filter above now?

bubble-wrap-style.yaml
bcamper: yesss

Cool to see this. But yes also makes sense to incur more cost server-side in the future where possible.

bubble-wrap-style.yaml
```diff
@@ -1382,11 +1431,111 @@ layers:
...
+#                      }
+#                   order: 1000
+#                   width: 1px
+#                   visible: true
+
+        hgv-interlay:
+            filter:
+                all:
+                    - hgv: true
+                    - function() { return global.sdk_hgv_overlay; }
```
bcamper, line 1522: On the Tangram side, we could add more direct support for `globals` in layer filters in the future, which would avoid the function compile/call. I actually did some work on this but never merged it, could consider for v0.20:

https://github.com/tangrams/tangram/compare/filter-globals
https://github.com/tangrams/tangram/tree/filter-boolean

Those would support patterns like:

```
filter:
  all:
    - hgv: true
    - global.sdk_hgv_overlay # this is a straight value substitution evaluating to true/false
```

```
filter:
  hgv: true
  $global.sdk_hgv_overlay # special $global syntax needed here, otherwise `global` will be interpreted as a feature property
```

bubble-wrap-style.yaml
```diff
@@ -1382,11 +1431,111 @@ layers:
...
+            draw:
+                lines:
+                    visible: true
```
bcamper: Superfluous (defaults to true)

bubble-wrap-style.yaml
```diff
@@ -1382,11 +1431,111 @@ layers:
...
+                    color: blue
+                    width: 1.7px
+            z-none:
```
bcamper: `z-none` and `z-restricted` filters are the same, is one supposed to be a different value?

bubble-wrap-style.yaml
```diff
@@ -1382,11 +1431,111 @@ layers:
...
+                draw:
+                    lines:
+                        color: purple
+                        width: 2.5px
+                tunnel:
+                    filter: {is_tunnel: true, $zoom: {min: 13} }
+                    draw:
+                        lines:
+                            # color: [[16, [0.843,0.843,0.843]]]
+                            color: purple
```
bcamper, line 1539: As written, this draw block doesn't do anything since it just re-sets the color to `purple` (which should already be `purple` from parent...?

bubble-wrap-style.yaml
```diff
@@ -1118,18 +1197,23 @@ layers:
...
                     order: global.feature_order
-                    color: [[11,[0.753,0.820,0.835]],[12,[0.710,0.800,0.824]],[13,[0.886,0.937,0.988]]]
+                    color: [[12,[0.753,0.820,0.835]],[13,[0.710,0.800,0.824]],[14,[0.886,0.937,0.988]]]
                     width: [[8,0px],[10,0px],[11,0.75px],[12,1px],[13,1px],[14,2px],[15,3px],[16,4px],[17,5px]]
                     join: round
                     outline:
-                        color: [[11,[0.671,0.788,0.812]],[13,[0.710,0.800,0.824]],[14,[0.635,0.812,0.843]]]
-                        width: [[9,0px],[11,0px],[12,0px],[13,0.5px],[14,0.75px],[15,0.75px],[16,0.75px],[17,1px]]
+                        color: [[12,[0.671,0.788,0.812]],[14,[0.710,0.800,0.824]],[15,[0.635,0.812,0.843]]]
+                        width: [[9,0px],[11,0px],[13,0px],[14,0.5px],[15,0.75px],[16,0.75px],[17,1px]]
```
bcamper: Can drop the extra `0px` stops now (anything <= z13 will be 0px):

`width: [[13,0px],[14,0.5px],[15,0.75px],[16,0.75px],[17,1px]]`

bubble-wrap-style.yaml
```diff
@@ -1382,11 +1431,111 @@ layers:
...
+                    draw:
+                        mapzen_icon_library:
+                            # you need to match any custom shield to the vector tile `network` values
+                            sprite: |
+                                function() {
+                                    if( feature.hgv_restriction_shield_text ) {
+                                        return ('generic_shield-' + feature.hgv_restriction_shield_text.length + 'char');
+                                    }
+                                }
+                            size: [[7,[75%,75%]],[12,[100%,100%]],[15,[110%,110%]]]
```
bcamper, line 1553: Should only need one % since it's the same for width/height:

`size: [[7, 75%], [12, 100%], [15, 110%]]`

bubble-wrap-style.yaml
```diff
@@ -1382,11 +1431,111 @@ layers:
...
+                            priority: 1
+                            # Size is specified below, and limited to 5 char in another filter
+                            repeat_group: shields
+                            repeat_distance: [[8,90px],[10,40px],[13,30px]]
```
bcamper, line 1557: I had to check in the code if `repeat_distance` supports zoom stops.. it does! Supports functions too. :) But the docs don't reflect either (updating).

@matteblair @tallytalwar does ES support zoom stops and/or functions for `repeat_distance`? Let's file an issue for parity if not.

bubble-wrap-style.yaml
```diff
@@ -1382,11 +1431,111 @@ layers:
...
+                            color: purple
+                            placement_min_length_ratio:
+                                - [8, 0.1]
+                                - [9, 0.25]
+                                - [10, 0.25]
+                                - [11, 0.15]
+                                - [12, 1]
+                                - [13, 1.50]
+                                - [14, 2.0]
+                            #cull_from_tile: true
```
bcamper: This is a deprecated JS keyword, we can remove entirely.

bubble-wrap-style.yaml
```diff
@@ -1382,11 +1431,111 @@ layers:
...
+                                        return ('generic_shield-' + feature.hgv_restriction_shield_text.length + 'char');
+                                    }
+                                }
+                            size: [[7,[75%,75%]],[12,[100%,100%]],[15,[110%,110%]]]
+                            priority: 1
+                            # Size is specified below, and limited to 5 char in another filter
+                            repeat_group: shields
+                            repeat_distance: [[8,90px],[10,40px],[13,30px]]
+                            placement: midpoint
+                            color: purple
```
bcamper, line 1559: May be superfluous depending on parent layer/filter questions above.

bubble-wrap-style.yaml
```diff
@@ -1382,11 +1431,111 @@ layers:
...
+                            placement_min_length_ratio:
+                                - [8, 0.1]
+                                - [9, 0.25]
+                                - [10, 0.25]
+                                - [11, 0.15]
+                                - [12, 1]
+                                - [13, 1.50]
+                                - [14, 2.0]
+                            #cull_from_tile: true
+                            visible: true
```
bcamper, line 1568: Superfluous as parent layers currently written.

bubble-wrap-style.yaml
```diff
@@ -1493,19 +1642,15 @@ layers:
                     $zoom: [13,14]
                 draw:
                     text-blend-order:
-                        priority: 50
```
bcamper, line 1496: ✂️ ❤️

bubble-wrap-style.yaml
```diff
@@ -2265,7 +2390,6 @@ layers:
                         - kind_detail: motorway
                 draw:
                     mapzen_icon_library:
-                        priority: 47
                         #color: yellow
```
bcamper: Not part of the diff but.. not needed :)

bubble-wrap-style.yaml
```diff
@@ -2389,9 +2506,8 @@ layers:
                         }
                     }
                 draw:
-                    mapzen_icon_library:
+                    mapzen_icon_library: {}
```
bcamper, line 2563: What's the deal with these lines (here and below) setting empty an empty object for the `draw` block? Can these just be removed or is that doing something? I'd expect it would just be a no-op when the layers are merged together.

bubble-wrap-style.yaml
```diff
@@ -2407,9 +2523,8 @@ layers:
...
```
bcamper, line 2580: See comment above on similar.

bubble-wrap-style.yaml
```diff
@@ -2490,6 +2603,13 @@ layers:
                     outline:
                         color: [0.799,0.799,0.799]
                         width: [[14, 0], [16, 0.5px], [18, 3px]]
+            construction:
+                filter: { kind: construction }
+                draw:
+                    dashed:
```
bcamper, line 2663: We *can* move the `dash` inline into the `draw` groups now, but probably outside (already large) scope of this PR. But for future, if we want to do that instead of having a separate style.

bubble-wrap-style.yaml
```diff
@@ -3154,13 +3275,14 @@ layers:
         draw:
             lines:
                 order: global.feature_order
-#                color: red
-#                width: [[9, 1px], [14, 2px], [16, 3px], [17, 8m]]
+                #debug:
```
bcamper, line 3332: Is keeping this useful? (Fine if so.)

bubble-wrap-style.yaml
bcamper: Interesting -- out of curiosity, what's a test case for this?

bubble-wrap-style.yaml
bcamper: huh, cool

bubble-wrap-style.yaml
```diff
@@ -3578,328 +3745,302 @@ layers:
...
+                    filter:
+                        all:
+                            - global.ux_point_of_view_kind_filter_country_capital
+                            - $zoom: { min: 5 }
                     draw:
                         mapzen_icon_library:
                             size: [[10,8px],[11,0px]]
                             sprite: capital-l
-                            priority: 5
-                        text-blend-order:
```
bcamper, line 3597: Was the `text-blend-order` removed because these are handled by the icon drawing?

bubble-wrap-style.yaml
```diff
@@ -4021,6 +4161,14 @@ layers:
                 mapzen_icon_library:
                     sprite: function() { return feature.religion; }
                     sprite_default: place_of_worship
+                    
+        generator:
+            filter: { kind: generator }
+            draw:
+                mapzen_icon_library:
+                    sprite: function() { return 'generator_' & feature.kind_detail; }
```
bcamper: I think you want `'generator_' + feature.kind_detail` here. You don't want `'generator_' & feature.kind_detail` (single `&` is a bitwise numeric operator) or `'generator_' && feature.kind_detail` (which would always return just `generator_`).

bubble-wrap-style.yaml
```diff
@@ -4041,293 +4189,6 @@ layers:
                         visible: false
         #end different icon sprites
 
-        #most POI icons without name should be hidden, except these
```
bcamper, line 4044: Ah is this tied to the `text-blend-order` removals above? (Or maybe I've lost track.)

bubble-wrap-style.yaml
```diff
@@ -4414,13 +4274,14 @@ layers:
...
             draw:
                 mapzen_icon_library:
-                    priority: 43
                     visible: false
-                    text:
-                        text_wrap: 10
-                        font:
-                            style: italic
-                            weight: normal
+                text:
```
bcamper: Just to confirm, this `text` is supposed to be an alternative to a `mapzen_icon_library` icon, which is turned off? Elsewhere we use that designated `text-blend-order` style for such cases, right?

bubble-wrap-style.yaml
```diff
@@ -4498,6 +4350,11 @@ layers:
                 draw:
                     mapzen_icon_library:
                         visible: false
+            low-priority-early-z14-icons:
+                filter: { kind_tile_rank: { min: 50 }, $zoom: [13,14] }
```
bcamper, line 4403: Interesting, haven't seen that `kind_tile_rank` in action before.

bubble-wrap-style.yaml
```diff
@@ -4802,6 +4664,7 @@ layers:
                 draw:
                     dots:
                         color: [[4,[0.525,0.765,0.686]],[9,[0.416,0.737,0.631]]]
+        # TODO: This seems broken
```
bcamper, line 4716: Tile content, or something I can help with?

bubble-wrap-style.yaml
```diff
@@ -5120,6 +5036,19 @@ layers:
                     color: [0.765,0.765,0.765]
                     visible: true
 
+#        # (20190729: nvk) TODO: This should only show when no other kind condition
```
bcamper, line 5088: I believe this kind of "else" condition can be done with the new `exclusive` layer property added in JS -- make all the other layers `exclusive`, and if none matches, the non-exclusive "else" one does -- but it's not in ES yet.

index.html
```diff
@@ -22,12 +22,13 @@
...
+        <!-- <script src="https://unpkg.com/tangram@latest/dist/tangram.min.js"></script> -->
         <!-- production nextzen -->
-        <!-- <script src="https://www.nextzen.org/tangram/0.16/tangram.min.js"></script> -->
+        <script src="https://www.nextzen.org/tangram/0.19/tangram.min.js"></script>
         <!-- production debug -->
-        <!-- <script src="https://mapzen.com/tangram/0.16/tangram.debug.js"></script> -->
+        <!-- <script src="https://mapzen.com/tangram/0.19/tangram.debug.js"></script> -->
         <!-- dev -->
+        <!-- <script src="https://line-label-repeats--tangram.netlify.com/dist/tangram.debug.js"></script> -->
         <!-- <script src="https://precog.mapzen.com/tangrams/tangram/master/dist/tangram.debug.js"></script> -->
```
bcamper, line 32: going wayyy back ;)

index.html
```diff
@@ -393,6 +394,55 @@
...
+                'Spain': 'es',
+                'Sweden': 'se',
+                'Taiwan': 'tw',
+                'Turkey': 'tr',
+                'Ukraine': 'ua',
+                'United Kingdom': 'gb',
+                'United States': 'us',
+                'Vietnam': 'vn'
+            };
+            // use query language, else default to English
```
bcamper: comment isn't relevant, from old copy-paste

bcamper (commented):
Great new features and I love all the code removal!

There are a few questions around filter logic or values that are worth addressing, may be one or two unintentional items, but otherwise all looks good.

I'd like to pull out a couple of these oft-repeated lines into globals, to be more concise and lower risk of copy-paste errors, such as:

`var kind = (global.ux_point_of_view && feature['kind:'+global.ux_point_of_view]) || (global.ux_point_of_view_fallback && feature['kind:'+global.ux_point_of_view_fallback]) || feature.kind;`

and

```
priority: |
                    function() { return (feature['min_zoom'] + (1 - 1 / feature['collision_rank']) * 0.1) || 65; }
```

I can take a crack at that.

bcamper:
OK I extracted some repetitive logic in:

f591907cbf414cab6beb4538295b53b92ff1705c
91f3b8f294d41bb124942474bf935b6a002b9a6b

Noticing that the pre-existing language toggle stuff could benefit from this as well, but it looks more complicated because of the many variants (short name, left/right borders, etc.), so I'll leave that for now.

bcamper:
Actually now that I'm looking at these lines like:

```
priority: |
  function() { return (feature['min_zoom'] + (1 - 1 / feature['collision_rank']) * 0.1) || 65; }
```

I'm realizing the default value of 65 will never be triggered. If `collision_rank` is missing, you'll get `-Infinity`. If `min_zoom` is missing, you'll still get a number, but less than `1`.

What's the desired fallback logic here?

bcamper:
So is this the desired collision priority logic:

- If both `min_zoom` and `collision_rank` exist, use the formula.
- If only `min_zoom` exists, use it (or use `min_zoom + 1` if features without a `collision_rank` should rank lower by default than those that do).
- If neither exist, use the provided default.

But, even then, the fallback values in the style like 65, 58, etc. don't make sense as final priority values when compared with `min_zoom` based values that are going to be in the ~1-20 range.

Are those numbers supposed to be defaults for `collision_rank`?

index.html
nvkelso: Fixed via d7defa5ef35a5bd1dae1e9d3a79da5b602715337.

bubble-wrap-style.yaml
nvkelso: Unrecognized country boundaries in Morocco, Israel, Kashmir areas.

Unrecognized region boundaries in Kashmir area.

bubble-wrap-style.yaml
nvkelso: The parent filter still needs to allow those 2 kinds... and then the sibling layer (right below this one) targets them for a custom style.

bubble-wrap-style.yaml
nvkelso: Supporting globals in layer tests directly instead of via function would be stellar :)

bubble-wrap-style.yaml
nvkelso: Fixed in 68bc27fb5a5056d7ed7b63d3b7660477fc4c135a.

bubble-wrap-style.yaml
nvkelso: That's a comment from a while ago, not sure what's going on there without some debugging that's beyond this PR.

bubble-wrap-style.yaml
nvkelso: YES! `exclusive` would be great here :) So in Tangram ES the layer won't match and won't display? Seems backwards compatible and okay to implement?

nvkelso: Reading the docs in https://github.com/tangrams/tangram/pull/705 I'm confused how I'd implement this. Is there a `!exclusive` or `priority: -1` to mean evaluate this one layer last? Or do I have to add priority to every other sibling layer here, too?

nvkelso:
You're right... those fallback values are a holdover from before all the `collision_rank` were moved server side and spread out with much greater detail. But something like this is still desired. If a feature has a `min_zoom` but no `collision_rank` then it should be evaluated as _worst_ among peers of the same `min_zoom`.

Example:

- Feature A:
    - `min_zoom`: 10
    - `collision_rank`: 1000
- Feature B:
    - `min_zoom`: 10
    - `collision_rank`: 1001
- Feature C:
    - `min_zoom`: 10
    - `collision_rank`: null

The outcome should be Feature **A** wins over **B** wins over **C**.

I think this could be accomplished by falling back to `feature.min_zoom + 0.999` and call it good?

nvkelso:
I've addressed the collision priority changes primarily in 92afcabb466831752f87ca73d17668dd5b57df8f (simplified to remove the defaults), and added logic for **road**, **water**, and **transit** labels (turn transit overlay on to see it in action).

nvkelso:
@bcamper This is ready for another review, please. I've addressed the collision priority stuff you called out (and fixed the transit overlay labels in the process), but don't know how to proceed on the https://github.com/tangrams/bubble-wrap/pull/277#discussion_r311863310 so maybe we defer that to a later PR?

bubble-wrap-style.yaml
```diff
@@ -5392,28 +5411,125 @@ layers:
...
+    transit-bus-roads:
+        data: { source: mapzen, layer: roads }
+        filter: { is_bus_route: true }
+        draw:
+            lines:
+                visible: global.sdk_transit_overlay
+                interactive: false
+                color: [[13,[0,0,0.5,0.25]],[15,[0,0,0.5,0.8]]]
+                width: [[11,0.5px],[12,0.8px],[16,1.5px],[18,1.5m]]
+                # let roads sort themselves past zoom 14
```
bcamper, line 5452: This is just assigning an explicit order value though, is the comment incorrect? (or the code?)

bubble-wrap-style.yaml
```diff
@@ -5392,28 +5411,125 @@ layers:
...
+                            style: italic
+                            size: 14px
+                            stroke: { color: white, width: 3px }
 
     transit-overlay-station-labels:
         data: { source: mapzen, layer: pois }
         filter:
             kind: [station, tram_stop, bus_station, subway_entrance, halt, stop, platform, bus_stop, stop_position ]
             any:
                 - area: false
```
bcamper: What's going on with both `area: false` and `area: true`?

bubble-wrap-style.yaml
bcamper: Still wondering about this one.


NVK: Removed in 930a049bc25e74953cfdf0e9bf4ca5d04dd8ea94.

bcamper (approved):
Made an additional fix to the those collision priority functions -- they need to check for `null` first, otherwise they'll divide by `null` which will be `Infinity` instead of falling back to the desired value.

Couple other small questions about things that may have issues, but probably not significant/blocking (please do take a look though).

bcamper:
> @bcamper This is ready for another review, please. I've addressed the collision priority stuff you called out (and fixed the transit overlay labels in the process), but don't know how to proceed on the [#277 (comment)](https://github.com/tangrams/bubble-wrap/pull/277#discussion_r311863310) so maybe we defer that to a later PR?

Agree we can follow up on this later.

bcamper:
> You're right... those fallback values are a holdover from before all the `collision_rank` were moved server side and spread out with much greater detail. But something like this is still desired. If a feature has a `min_zoom` but no `collision_rank` then it should be evaluated as _worst_ among peers of the same `min_zoom`.
> ...
> I think this could be accomplished by falling back to `feature.min_zoom + 0.999` and call it good?

Yep that works, made some updates in https://github.com/tangrams/bubble-wrap/pull/277/commits/8fdfe3be79e87e5bb48e355a7f30c19c5ba14a3d.

bubble-wrap-style.yaml
nvkelso: Updated the comment in 1da2c657047b2b9d6ba99306a5fa0ee6010f466e.

We moved the zoom changes server side a few releases ago so the comment was outdated. There look to be a few static int values in the scene file still. I've filed https://github.com/tangrams/bubble-wrap/issues/278 to track that.

bubble-wrap-style.yaml
nvkelso: That's odd, probably a holdout from around v1.0 tiles? 

In any event, removing it doesn't break anything and simplifies the style: eafad2608ad9a4ba8510daabd71556d99c504928.

nvkelso (commented):
I fixed a noticeable regression in the style between prod and this PR by changing order of operations in the `collision_priority` function in eafad2608ad9a4ba8510daabd71556d99c504928, which was preventing neighbourhood labels in SF to show up at zooms less than 16 and was preventing Golden Gate Park label from showing up w/r/t roads labels, and too many "bay" labels north of San Francisco.

bubble-wrap-style.yaml
nvkelso: This was needed because it "folds" in with the general roads layer styling which has a more specific selector for road tunnels which reset the purple coloring in this layer's parent. Fun times.

bubble-wrap-style.yaml
nvkelso: See: https://github.com/tangrams/tangram-es/issues/2079.

bubble-wrap-style.yaml
nvkelso: I set to `draw: mapzen_icon_library: {}`  because it's not obvious to me if I can have a sublayer of a parent layer without a draw block (next few lines down). So it was just me being caution & clever at the same time.

bubble-wrap-style.yaml
nvkelso: Agree... though Tangram ES not supporting that kinda puts a damper on that / map design is broken without the dashing.

bubble-wrap-style.yaml
nvkelso: Yes... I keep some debugging in when I keep going back to debug (see buildings and roads).

bubble-wrap-style.yaml
nvkelso: It's because the master layer's default `text-blend-order` sets `priority: global.collision_priority` above, so they aren't needed every instance – since all the specific values moved server side and it's just a lookup now.

bubble-wrap-style.yaml
nvkelso: This was all moved server side, and the diff here makes that hard to read.

bubble-wrap-style.yaml
bcamper: Oh right, forgot it's not in ES yet.

bubble-wrap-style.yaml
nvkelso: Well, funny you should ask. If I move to `text-blend-order` instead of `text` around 50% of the labels are culled when I expect them to be shown :\

http://localhost:8000/#18/37.73212/-122.50216

image

nvkelso: Well, I don't know why but this other suggestion of yours is the root cause of the missing text:

>-                    size: [[7,[75%,75%]],[12,[100%,100%]],[15,[110%,110%]]]
>+                    size: [[7,[75%]],[12,[100%]],[15,[110%]]]

PR #279, opened 2021-11-29
switch ref to network for orange roads route colouring

nvkelso:
There should be no visual change, but as properties are stripped out at low zooms, we should use the reliable Tilezen ones instead of the random OSM ones per https://github.com/tilezen/vector-datasource/pull/2008 and https://github.com/tilezen/vector-datasource/issues/1998.

I switch this to `network` instead of `shield_text` because at some zooms `shield_text` is dropped to reduce tile size and increase feature merging.

SF: 12/37.7582/-122.2903

before
Screen Shot 2021-11-28 at 21 07 06

after
Screen Shot 2021-11-28 at 21 07 15

London: 11/51.4809/-0.0778

before
Screen Shot 2021-11-28 at 21 08 08

after
Screen Shot 2021-11-28 at 21 08 26

### DIFF
--- a/bubble-wrap-style.yaml
+++ b/bubble-wrap-style.yaml
@@ -1184,14 +1184,14 @@ layers:
             grid:
                 order: global.feature_order
                 #color: global.water1
-            
+
         ocean:
             filter:
                 kind: ocean
             draw:
                 grid:
                     color: global.water1
-                    
+
         inland-water:
             filter:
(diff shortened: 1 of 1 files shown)


nvkelso:
I switch this to `network` instead of `shield_text` because at some zooms `shield_text` is dropped to reduce tile size and increase feature merging.

